PR #2, opened 2015-04-02
Serial leaning

mfine:
Replaces https://github.com/swift-nav/piksi_firmware/pull/306 and fixes https://github.com/swift-nav/piksi_firmware/issues/295.

Another pass at weaning off dependencies in the script and updating them. The approach here is to move the functionality of `sbp.client.main` into `piksi_tools.serial_link` and continue to use it as is. Moved over all of the scripts and hand ran them - working well, including `serial_link` and `bootloader`.

Console is mostly done, but setup of console needs some work.

/cc @mookerji


piksi_tools/testing/print_device_details.py
```diff
@@ -25,43 +25,43 @@
...
 
-for k, v in sv.settings.iteritems():
-  if k != 'system_info':
-    print "%s:" % k
-    for k_, v_ in v.iteritems():
-      print "    %-20s %s" % (k_, v_)
+    print
+    print "Settings"
+    print "--------"
     print
```
mookerji: This device details thing is pretty useful data dumped whenever we do parsing. The content type (sectioned-plain text) is hard to parse, but it would be cool if we had the option to output json, yaml, etc.

piksi_tools/serial_link.py
```diff
@@ -9,447 +9,162 @@
...
-    self.timer.daemon = True
-    self.timer.start()
+  sys.stdout.write(sbp_msg.payload)
 
-def default_watchdog_alarm():
+def main():
   """
-  Called when the watchdog timer alarms. Will raise a KeyboardInterrupt to the
-  main thread and exit the process.
+  Get configuration, get driver, get logger, and build handler and start it.
```
mookerji: so the actual serial link command line tool's going to continue to live here instead of at: https://github.com/swift-nav/libsbp/blob/master/python/sbp/client/main.py. In that case, we'll probably have to clarify with "libsbp has the dev libraries" and "piksi_tools has the actual command line tool" or something.

mfine: Yes. All these scripts felt more like tools and were heavily integrated with the console and the like - don't know, made less sense to have it in sbp. Will augment the examples with a similar like script, but keep the main artifact in piksi_tools.

mfine:
Also contemplating moving the console stuff (views, setup script) into a `piksi_tools.console` to de-clutter things here.

mfine:
Have a working console! The console is still pretty janky, but this should be a good baseline to improve on.

mfine:
Struggling to get the setup.py to get all the package data incorporated. Yuck. Can't find images and settings.yaml once these packages get installed. And maybe we don't want the console run out of pip (other dependencies are not being tracked (qt4, etc.)). Anyway, things to iterate on here.

mfine:
This should be all set to bring in and we should start thinking about transitioning to this formally.

piksi_tools/console.py
```diff
@@ -9,49 +9,71 @@
...
-args = parser.parse_args()
-serial_port = args.port[0]
+import sbp.client.handler
+from sbp.logging import SBP_MSG_DEBUG_VAR, SBP_MSG_PRINT
+from sbp.piksi import SBP_MSG_RESET
+from sbp.client.drivers.pyserial_driver import PySerialDriver
+from sbp.client.drivers.pyftdi_driver import PyFTDIDriver
+from piksi_tools import __version__ as CONSOLE_VERSION
+
+# Shut chaco up for now
```
mookerji: HA

mookerji:
lgtm!

@fnoble @cbeighley you might want to check this PR out.

PR #8, opened 2015-04-07
add argument to settings view

denniszollo:


mookerji:
@denniszollo you may want to rebase this off of mark's new PR for the console after it gets merged.

denniszollo:
@mfine LMK what you think of this fix for issue .  I should have never committed that hardcoded nonsense.  I had a corresponding PR for piksi_firmware but am dropping it for now as we should be abandoning the console there shortly.

piksi_tools/console/settings_list.py
```diff
@@ -9,36 +32,55 @@ def __str__(self):
...
+      stram = open(path_to_file, "r")
+      self.list_of_dicts = yaml.load(stram)
+      # inform user of success or failure
+      print "Loaded settings yaml file from path " + path_to_file
+      print "Number of settings loaded {0}".format(len(self.list_of_dicts))
+    except IOError as e:
+      print "I/O error({0}): {1} with file {2}".format(e.errno, e.strerror,
+                                                       path_to_file)
+    except:
+      print "Unexpected error:", sys.exc_info()[0]
```
mfine: The more common pattern here seems to be:

```python
   except:
      import traceback
      traceback.print_exc()
```

piksi_tools/console/settings_list.py
```diff
@@ -9,36 +32,55 @@ def __str__(self):
...
   def return_groups(self):
-    group_set = set()
     output = []
-    for element in self.list_of_dicts:
-        group = element['group']
-        if group not in group_set:
-          output.append(element['group'])
-          group_set.add(group)
+    #confirm our class actually has list_of_dicts
+    if len(self.list_of_dicts) != 0 :
```
mfine: Instead of always checking if the length of the dicts is not zero, how about initializing `list_of_dicts` to `{}`, and then updating it if you're able to load the yaml successfully.

mfine: This only applies to areas where you're going to walk the list next (as in this example). And a more idiomatic way to check if a list is empty (vs getting its length, I think is linear for non-empy lists) is to do:

```python
if not self.list_of_dicts:
```

denniszollo:
Thanks for the feedback @mfine .  Little by little I will behave in a Pythonic fashion.

piksi_tools/console/settings_view.py
```diff
@@ -287,7 +287,9 @@ def set(self, section, name, value):
       self.link.send(SBP_MSG_SETTINGS,
           '%s\0%s\0%s\0' % (section, name, value))
 
-  def __init__(self, link, read_finished_functions=[], gui_mode=True):
+  def __init__(self, link, read_finished_functions=[],
+               name_of_yaml_file="settings.yaml", gui_mode=True):
```
denniszollo: I may add an arg to the console.py to maybe specify the path to be used in the constructor

piksi_tools/console/settings_list.py
```diff
@@ -1,44 +1,84 @@
...
+        root = __file__
+        if os.path.islink (root):
+            root = os.path.realpath (root)
+        return os.path.dirname (os.path.abspath (root))
+    except:
+        print "There is no __file__ variable. Please contact the author."
 
 class SettingsList():
-  list_of_dicts = list();
+  list_of_dicts = {}
```
mfine: Actually, think this was ok, since it was a list (my bad) - should just be initialized to `list()` without a semi-colon.

piksi_tools/console/settings_list.py
```diff
@@ -1,44 +1,84 @@
...
   def __str__(self):
     pp = pprint.PrettyPrinter(indent=2)
     pp.pprint(self.list_of_dicts)
 
   def get_dict(self, group, name):
-    for element in self.list_of_dicts:
-        if element['name'] == name and element['group'] == group:
-          return element
+    # confirm our class actually has list_of_dicts
+    if self.list_of_dicts :
```
mfine: You shouldn't need this since here - since you're walking the list below, its a spurious check.

mfine:
Looks good!

PR #13, opened 2015-04-10
Serial cleanup and console relocation

### LINKED ISSUE #10: Move the console into piksi_tools.console
The console clutters up the main tools area - its hard disambiguating what's a script vs what's part of the console (views, etc.) plus all the installer stuff. What about moving the console into its own directory? `piksi_tools.console`. Might help think about whats in the tools directory better.

/cc @fnoble @denniszollo @mookerji

mfine:
Includes the changes from #2 and moves everything console-related into its own package to fix #10.

/cc @mookerji @fnoble


mookerji:
.... so this looks good to me. I think that we should probably setup a console build next week to point to this repo and do some broader play testing. We could also maybe run serial_link in the HITL testing setup from here as well.

mfine:
Sounds good - I'm going to bring this in. It's not active anywhere, so should be safe and not damaging. We can then work on transitioning here once we feel good about the quality.

PR #22, opened 2015-04-15
Add 'system_monitor.watchdog' setting to enable/disable watchdog timer.

henryhallam:


mfine:
Settings look good to me - @denniszollo?

fnoble:
I wouldn't worry about putting the firmware version in there, if the setting doesn't exist in the firmware then it won't show up

The list of settings is stored on the device - the YAML is just to cross reference the descriptions.

henryhallam:
@fnoble Oh, good point.
Incidentally, for the ext_events should I bump the required NAP revision?  It's currently ancient, 0.9something. Or is it better to not require users to upgrade their NAP, and just leave the feature nonfunctional?

fnoble:
I would bump the NAP version if the FW won't work without it. I'd leave it if it just won't detect events.

PR #23, opened 2015-04-15
Support logging JSON and pickle at the same time

mfine:
This is an outstanding capability not supported by `piksi_tools`'s `serial_link.py`. Before moving on to `piksi_tools`, we need to be able to write JSON and pickle logs at the same time. Kind of ugly - there should be a better way to nest these context, but for now it should be ok.

Once in, we can move Jenkins on to `piksi_tools`.

Towards supporting https://github.com/swift-nav/piksi_firmware/issues/349.

/cc @mookerji @fnoble

### DIFF
--- a/piksi_tools/serial_link.py
+++ b/piksi_tools/serial_link.py
@@ -167,34 +167,37 @@ def main():
     # Handler with context
     with Handler(driver.read, driver.write, args.verbose) as link:
       # Logger with context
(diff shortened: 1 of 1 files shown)


fnoble:
I'm down with doing it this way until we can kill pickle logs off entriely

mfine:
Fixed - now takes a -j, --json flag to use JSON.

PR #25, opened 2015-04-17
Pickle. It was a thing

mfine:
Remove pickle generation. Also remove the more exotic inputs and outputs - kind of clutters up this simple tool. They more belong in the examples area of sbp.

/cc @fnoble @mookerji


fnoble:
image

excellent

PR #26, opened 2015-04-18
create simple script for rewriting logs from pickle to json

denniszollo:

### DIFF
--- a/piksi_tools/pickle_to_json.py
+++ b/piksi_tools/pickle_to_json.py
@@ -1,0 +1,64 @@
+
+from sbp.client.loggers.json_logger import JSONLogIterator
+from sbp.client.loggers.pickle_logger import PickleLogIterator
+import json
+import struct
+import sbp
+import sbp.table
+import sys
+sys.modules['sbp_piksi'] = sbp  # creates a packageA entry in sys.modules
+
+def rewrite(datafiles, process_functions=[]):
+  """
+  Rewrite pickle file as a json file with a similar name
+  Parameters
+  ----------
+  datafiles : list
+      list of files on which to operate.
+  process_functions : list
+      list of functions to call on the message as it is converted
+  """
+  for f in datafiles:
+      with open(f + ".json", 'w+') as log_file:
+        with PickleLogIterator(f) as log:
+          for delta, timestamp, msg in log.next():
+            for process in process_functions:
+                msg = process(msg)
+            m = struct.pack("<HHB", msg.msg_type, msg.sender, len(msg.payload))
+            m += msg.payload
+            crc = sbp.crc16(m[1:])
+            msg.crc = crc
+            m = {"delta": delta,
+                 "timestamp": timestamp,
+                 "data": sbp.table.dispatch(msg).to_json_dict()}
+            log_file.write(json.dumps(m) + "\n")
+          print "Succesfully wrote json output to {0}".format(log_file)
+
+def translate_emphemeris(msg):
+  """
+  Handle legacy ephemeris messages
+  Parameters
+  ----------
+  msg : dictionary representing a SBP message
+  Returns
+  ----------
+  updated message
+  """
+  if msg.msg_type==26:
+    msg.msg_type = 70
+  return msg
+
+if __name__ == "__main__":
+  """
+  Main function for log translating.
+  """
+  import argparse
+  parser = argparse.ArgumentParser(description='Swift Nav SBP log converter.')
+  parser.add_argument('datafiles',
+                      nargs='+',
+                      help='Specify the file(s) to rewrite from pickle to json')
+  args = parser.parse_args()
+  datafiles = args.datafiles
+  rewrite(datafiles,[translate_emphemeris])
+
+


denniszollo:
Would you mind reviewing @mfine

piksi_tools/pickle_to_json.py
```diff
@@ -0,0 +1,52 @@
...
+from sbp_log_analysis.solution import SolutionTest
+from sbp_log_analysis.startup import StartupTest
+from sbp_log_analysis.test import SBPTestSuite
+from sbp_log_analysis.thread import ThreadTest
+from sbp_log_analysis.uart import UARTTest
+from sbp_log_analysis.utils import deep_eq
+from sbp_log_analysis.utils import deep_eq
+from sbp.client.loggers.json_logger import JSONLogIterator
+from sbp_log_analysis.tracking import TrackingTest
+from sbp.client.loggers.pickle_logger import PickleLogIterator
```
mfine: I don't think you need any of the above imports.

piksi_tools/pickle_to_json.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+if __name__ == "__main__":
+  import argparse
+  parser = argparse.ArgumentParser(description='Swift Nav SBP log converter.')
+  parser.add_argument('files',
+                      nargs='+',
+                      help='Specify the file(s) to rewrite from pickle to json')
+  args = parser.parse_args()
+  datafile = args.files
```
mfine: Maybe call this `datafiles`

piksi_tools/pickle_to_json.py
```diff
@@ -0,0 +1,52 @@
...
+  if datafile:
```
mfine: Don't think you need this check - it won't iterate if its a None value inside the for loop of the routine - either that or you can have rewrite work on a single file and loop from here - but not big difference.

piksi_tools/pickle_to_json.py
```diff
@@ -0,0 +1,52 @@
...
+
+def rewrite(files):
+  """ Rewrite pickle file as a json file with a similar name
+  """
+  for f in files:
+    with open(f, "r") as handle:
+      with open(f + ".json", 'w+') as log_file:
+        with PickleLogIterator(handle) as log:
+          for delta, timestamp, msg in log.next():
+            if msg.msg_type == 26:
```
mfine: Did you want to leave this mapping in? This is specfically for handling the old ephemeris message type cases.

mfine: Maybe to make this more general, there should be a function to pass in with this to capture the change you want to make to the message before you write it back out as JSON - and the old ephemeris message would be one use case of that.

piksi_tools/pickle_to_json.py
```diff
@@ -0,0 +1,64 @@
+
```
mookerji: nit: missing header

piksi_tools/pickle_to_json.py
```diff
@@ -0,0 +1,64 @@
...
+from sbp.client.loggers.json_logger import JSONLogIterator
+from sbp.client.loggers.pickle_logger import PickleLogIterator
+import json
+import struct
+import sbp
+import sbp.table
+import sys
+sys.modules['sbp_piksi'] = sbp  # creates a packageA entry in sys.modules
```
mookerji: nice

piksi_tools/pickle_to_json.py
```diff
@@ -0,0 +1,64 @@
...
+  process_functions : list
+      list of functions to call on the message as it is converted
+  """
+  for f in datafiles:
+      with open(f + ".json", 'w+') as log_file:
+        with PickleLogIterator(f) as log:
+          for delta, timestamp, msg in log.next():
+            for process in process_functions:
+                msg = process(msg)
+            m = struct.pack("<HHB", msg.msg_type, msg.sender, len(msg.payload))
```
mookerji: libsbp's SBP object has a pack function that should be used here.

piksi_tools/pickle_to_json.py
```diff
@@ -0,0 +1,64 @@
...
+            m += msg.payload
+            crc = sbp.crc16(m[1:])
+            msg.crc = crc
+            m = {"delta": delta,
+                 "timestamp": timestamp,
+                 "data": sbp.table.dispatch(msg).to_json_dict()}
+            log_file.write(json.dumps(m) + "\n")
+          print "Succesfully wrote json output to {0}".format(log_file)
+
+def translate_emphemeris(msg):
```
mookerji: nit: ephemeris is misspelled. 

Backwards compatibility is a hairy question, but this is fine for time being. I might generalize the application of `translate_ephemeris` to `migrate_msg` or something, and provide a list of rewrites that can be applied. This isn't the last case we're going to have something like this.

piksi_tools/pickle_to_json.py
```diff
@@ -0,0 +1,64 @@
...
+  import argparse
+  parser = argparse.ArgumentParser(description='Swift Nav SBP log converter.')
+  parser.add_argument('datafiles',
+                      nargs='+',
+                      help='Specify the file(s) to rewrite from pickle to json')
+  args = parser.parse_args()
+  datafiles = args.datafiles
+  rewrite(datafiles,[translate_emphemeris])
+
+
```
mookerji: A test would be great here, if possible.

denniszollo:
not really relevant anymore.  Closing.

PR #31, opened 2015-04-23
Add an append file option to serial_link.py

mfine:
Add an option to append to a log file at a location. Maybe one day this could be an "append" mode that we choose, but I'd like to run this concurrently with existing logging so that I can phase things in without disruption.

This change depends on https://github.com/swift-nav/libsbp/pull/132.

/cc @mookerji

### DIFF
--- a/piksi_tools/serial_link.py
+++ b/piksi_tools/serial_link.py
@@ -72,6 +72,9 @@ def get_args():
   parser.add_argument("-o", "--log-filename",
                       default=[LOG_FILENAME], nargs=1,
                       help="file to log output to.")
+  parser.add_argument("-a", "--append-log-filename",
+                      default=[None], nargs=1,
+                      help="file to append log output to.")
   return parser.parse_args()
 
 def get_driver(use_ftdi=False, port=SERIAL_PORT, baud=SERIAL_BAUD):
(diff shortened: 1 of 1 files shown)


piksi_tools/serial_link.py
```diff
@@ -107,6 +110,20 @@ def get_logger(use_log=False, filename=LOG_FILENAME):
   print "Logging at %s." % filename
   return JSONLogger(filename)
 
+def get_append_logger(filename):
```
mookerji: Have you looked into what manner of appender might work here? I've only looked at https://docs.python.org/2/library/logging.handlers.html#rotatingfilehandler and https://docs.python.org/2/library/logging.handlers.html#timedrotatingfilehandler, but don't know if there's a better option tied to fluentd.

mfine: In the past I've done it outside of the language with `logrotate` see [logrotate8](http://linuxcommand.org/man_pages/logrotate8.html).

mookerji: Great! I'm convinced.

piksi_tools/serial_link.py
```diff
@@ -143,33 +161,35 @@ def main():
...
-              time.sleep(0.1)
-            print "Timer expired!"
-        except KeyboardInterrupt:
-          # Callbacks, such as the watchdog timer on SBP_HEARTBEAT call
-          # thread.interrupt_main(), which throw a KeyboardInterrupt
-          # exception. To get the proper error condition, return exit code
-          # of 1. Note that the finally block does get caught since exit
-          # itself throws a SystemExit exception.
-          sys.exit(1)
+        with get_append_logger(append_log_filename) as append_logger:
```
mookerji: nit: from the API point of view, what about just having a file mode flag instead of a new argument?

mfine:
Bringing this in.

piksi_tools/serial_link.py
mfine: @mookerji definitely - in the comments for the PR I noted

```
Maybe one day this could be an "append" mode that we choose, but I'd like to run this concurrently with existing logging so that I can phase things in without disruption.
```

For transition purposes, I'd like to be able to do both - will fold it into a mode afterwards once we're moved over.

PR #43, opened 2015-05-07
Inject tags into logs

mfine:
Need to inject tags into logs - namely `session_id` and potentially other metadata. This completes https://github.com/swift-nav/libsbp/pull/145.

/cc @mookerji

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -50,6 +50,7 @@ coverage.xml
 
 # Django stuff:
 *.log
+*.log.json
 
 # Sphinx documentation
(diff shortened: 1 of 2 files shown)


mookerji:
Looks good. Is this going to require any piksi_tools/libsbp consumers to migrate to a new libsbp?

mfine:
To use the flag and start injecting tags into the logs, that will need a new libsbp - until then, other clients should be unimpacted until the new sbp comes in.

PR #44, opened 2015-05-07
add headers to csv files auto logged by console

denniszollo:


piksi_tools/console/baseline_view.py
```diff
@@ -180,6 +180,7 @@ def baseline_callback(self, sbp_msg):
 
       if self.log_file is None:
         self.log_file = open(time.strftime("baseline_log_%Y%m%d-%H%M%S.csv"), 'w')
+        self.log_file.write('time,meters_north,meters_east,meters_down,distance(meters),num_sats,flags,num_hypothesis')
```
fnoble: maybe change meters_north -> north(meters) to be consistent with the other fields (and same with east, down)?

fnoble:
Awesome! lgtm apart from that small tweak

fnoble:
Oh, wait, do you want to put a `#` on the front of that line? Whats the standard for CSV?

fnoble:
Actually, scratch that, seems like there is no real standard but most people don't comment the header line. Even how to do comments is not standardized :(

PR #50, opened 2015-05-14
don't clear the settings when you press the read button

denniszollo:
Removing the settings when you "press the read button" was causing print_device_details.py to fail due to timing issues in the HITL.  There is no reason to wipe out the existing settings as they will just be read again at startup or when the read button is fired.

### DIFF
--- a/piksi_tools/console/settings_view.py
+++ b/piksi_tools/console/settings_view.py
@@ -195,7 +195,6 @@ class SettingsView(HasTraits):
   )
 
   def _settings_read_button_fired(self):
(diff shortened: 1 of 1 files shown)


denniszollo:
@cbeighley Before you do more testing I would like to merge this.  I think it is the reason behind some of the failed capture jobs.

cbeighley:
lgtm

fnoble:
I feel like I did this for a reason, just looking into it

fnoble:
too late :(

PR #53, opened 2015-06-01
added reset button

denniszollo:
This adds a reset button to the "settings view" on the console.


denniszollo:
@cbeighley would you mind reviewing?

cbeighley:
lgtm...assuming that button looks decent

denniszollo:
image

It looks like this @cbeighley

fnoble:
@denniszollo : do you think it should be on the settings page? Might make more sense on the system monitor page potentially?

denniszollo:
@fnoble .  Yes I had a hard time deciding where to put this feature. If there is consensus about the system monitor Lets put it there.

Additionally, I sort of see a suite of features that don't fit into a specific tab which might make sense to have as a menu bar at the top, or down by the console log which include:
  *location / name of log files.  
  *Record files on/off
  *Piksi mode (RTK, float, etc)
  *Piksi status indicator
  *Reset Piksi

cbeighley:
+1 I think a status bar independent of the tabs would be good

denniszollo:
Made a new PR that adds button to the system Monitor view.

PR #54, opened 2015-06-02
Settings generation

denniszollo:
This resurrects the settings pdf generation in piksi_tools (where the settings.yaml lives)


denniszollo:
When we cut v0.17 I would like to put out a new settings.pdf as well, so It would be nice to resurrect this.

denniszollo:
@mookerji @mfine any issues with merging this in so we can generate settings.pdf again?

mfine:
If we bring this into piksi tools, which I'm excited about, then it should get buried somewhere with other tech debt we want to remove (like the stuff in `piksi_tools/testing`... like dump it in `piksi_tools/console/settings`). But ultimately this belongs under libsbp, not piksi_tools, if we want one place to generate our documentation from and we'd like to incorporate this documentation into the sbp document.

denniszollo:
ok lets put get it into libsbp then for real! I assume that also means settings.yaml will go into libsbp.

mookerji:
Hmmmmmm, what about just putting it into its own repo? I used to be in the settings-should-be-in-libsbp crowd, but then, it's actually a wholly separate concept not related to tools for using the Piksi's messaging.

fnoble:
I'm also in the this is (logically) separate from SBP crowd although I don't care that much which repo it ends up in. The way I see it is SBP is a protocol, i.e. an interface between devices that is largely agnostic to what those devices are. The settings stuff is a description of the piksi firmware settings which is very device specific. They are only really related by the fact that the settings are exposed over SBP (which SBP supports in a fairly non device specific way).

mfine:
It seems like the settings line up with content that's actually captured in libsbp:

```
bootload.yaml
file_io.yaml
flash.yaml
piksi.yaml
settings.yaml
system.yaml
```

Aren't these settings descriptions just an expansion of the description of the possible settings messages? I agree that these setting descriptions are piksi-specific, but that's also true of a large portions of the described messages in sbp.

mookerji:
That's a fair point, actually.

There's a PR from yesteryear where I put in the settings.yaml for libsbp, and you maybe able to just get it out of there.

denniszollo:
I'd really like to decide on our course of action for this before the 1.0 - do we have consensus on putting the generation and resulting PDF document in libsbp?

denniszollo:
Current Plan (not to be implemented right now, but at least it's planned.)  
1. add settings_annotation.yaml to the spec folder
2. move generate_piksi_settings.py to the spbg: this script will do 3 things
   a. make a standalone piksi_settings.pdf
   b. make a real python dictionary of dictionaries to be in SBP (piksi_settings.py).  From this dict of you can specify the compound key (group plus setting name) and get metadata.  You include this from sbp/piksi/setting or something
  c. make a tex file that can be added to the SBP.pdf doc
3. this generate_piksi_settings.py will be called by makefile
4. Add a "this document has been moved message" to the docs/settings.pdf in piksi firmware pointing to the SBP version of the document.

PR #55, opened 2015-06-08
Bring versioning back in

mfine:
Versioning by hand has gotten old. Pulled the old versioning script back in - tweaked it to handle versioning with leading `v`. Have to replicate the logic in setup.py because setup.py can't take a dependency on library code - that was the initial issue we had with that approach. Identical code was added to https://github.com/swift-nav/libsbp/pull/154 - will take comments for both here.

/cc @fnoble @mookerji

### DIFF
--- a/MANIFEST.IN
+++ b/MANIFEST.IN
@@ -9,7 +9,6 @@ include .gitignore
 include LICENSE
 include tox.ini
 include *.yaml
-include piksi_tools/RELEASE-VERSION
 recursive-include piksi_tools/ *.py
 recursive-include piksi_tools/console/images *
 prune docs/_build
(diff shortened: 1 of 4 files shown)


mookerji:
Regarding the duplicate file believe that `version.py` was previously in the top-level with `setup.py`, where it was accessible for import to setup.py.

mfine:
But I think you want the version available under sbp so that dependencies can access it - I think if it's at the top level, then it's not accessible. We could move this into a verision.py, but I still think we're going to need two of these files, given my understanding of how python packaging works.

mfine:
Removed the duplicate versions.

PR #57, opened 2015-06-08
Versioned bootloader

mfine:
Brought in the new SBP and setup new handler for new bootloader message. Not actually using `sbp_protocol` yet. Not clear exactly where explicitly it needs to be used. Still using the bootloader version everywhere instead.

/cc @cbeighley

### DIFF
--- a/piksi_tools/bootload.py
+++ b/piksi_tools/bootload.py
@@ -43,7 +43,8 @@ class Bootloader():
     self.stopped = False
     self.handshake_received = False
     self.version = None
-    self.sbp_protocol = None
+    # SBP version is unset in older devices.
+    self.sbp_version = 0
     self.link = link
     self.link.add_callback(self._deprecated_callback, SBP_MSG_BOOTLOADER_HANDSHAKE_DEPRECATED)
     self.link.add_callback(self._handshake_callback, SBP_MSG_BOOTLOADER_HANDSHAKE_DEVICE)
@@ -71,8 +72,7 @@ class Bootloader():
 
   def _handshake_callback(self, sbp_msg):
     self.version = sbp_msg.version
-    # Taking the sbp_protocol here, but continuing to use the bootloader version below.
-    self.sbp_protocol = sbp_msg.flags & 0xff
+    self.sbp_version = sbp_msg.flags & 0xff
     self.handshake_received = True
 
   def wait_for_handshake(self, timeout=None):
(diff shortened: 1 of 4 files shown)


mfine:
@cbeighley moved all of the logic to be driven by the `sbp_version`.

piksi_tools/bootload.py
```diff
@@ -78,9 +87,9 @@ def wait_for_handshake(self, timeout=None):
     return True
 
   def reply_handshake(self):
-    # < v2.0 of the bootloader, reuse single handshake message.
-    if self.version < "v2.0":
-      self.link.send(SBP_MSG_BOOTLOADER_HANDSHAKE_DEVICE, '\x00')
+    # < 0.45 of SBP protocol, reuse single handshake message.
+    if self.sbp_version < 45:
```
cbeighley: Should sbp_version be compared as a float or version string instead? Integer minor version seems odd.

piksi_tools/bootload.py
```diff
@@ -156,12 +165,14 @@ def main():
...
         print "Piksi Onboard Bootloader Version:", piksi_bootloader.version
+        if piksi_bootloader.sbp_version > 0:
+          print "Piksi Onboard SBP Protocol Version:", piksi_bootloader.sbp_version
 
         # Catch all other errors and exit cleanly.
         try:
           import flash
           with flash.Flash(link, flash_type=("STM" if use_stm else "M25"),
-                           version=piksi_bootloader.version) as piksi_flash:
+                           sbp_version=piksi_bootloader.sbp_version) as piksi_flash:
```
cbeighley: This is super clear where the versioning is coming from now.

piksi_tools/bootload.py
mfine: I was comparing it as a 16-bit value, without pulling out the 8-bit minor and 8-bit major. Could be confusing once we rev? E.g., 1.1 = 257.

piksi_tools/stm_unique_id.py
```diff
@@ -19,22 +19,31 @@
...
-  def __init__(self, link, version):
+  def __init__(self, link):
+    self.heartbeat_received = False
     self.unique_id_returned = False
+    # SBP version is unset in older devices.
+    self.sbp_version = 0
     self.unique_id = None
     self.link = link
-    self.version = version
+    link.add_callback(self.receive_heartbeat, SBP_MSG_HEARTBEAT)
```
cbeighley: Oh nice. I was thinking it would get passed when initialized after receiving the heartbeat but this is way nicer.

piksi_tools/bootload.py
cbeighley: Yeah I agree it would be confusing.

mfine: Moved the version to a tuple `(0, 45)`.

cbeighley:
lgtm

PR #58, opened 2015-06-09
add -o command line arg

denniszollo:
This will allow you to put a custom prefix on a logfile (will aid in testing with the console)


piksi_tools/console/console.py
```diff
@@ -312,7 +317,8 @@ def __init__(self):
 
 with serial_link.get_driver(args.ftdi, port, baud) as driver:
   with sbp.client.handler.Handler(driver.read, driver.write, args.verbose) as link:
-    with serial_link.get_logger(args.log, serial_link.LOG_FILENAME) as logger:
+    logfilename = log_file_prefix + "_" + time.strftime("%Y%m%d-%H%M%S.log.json")
```
mookerji: This should be probably be `logfilename = log_file_prefix + "_" + `serial_link.LOG_FILENAME)`

mookerji: Or even better, just having the parser argument above be exactly: 

https://github.com/swift-nav/piksi_tools/blob/3fcb4143c0ce3143f1efe311327798873755fdc4/piksi_tools/serial_link.py#L72

so that things will be consistent. Then you can pass in the date, prefix, etc. from the command line.

denniszollo: The goal is to make this feature accessible to an end user.  Putting together a date at the command line is pretty difficult and adds overhead.  What's wrong with appending the date?

piksi_tools/console/console.py
```diff
@@ -25,6 +25,7 @@
 
 # Shut chaco up for now
 import warnings
+import time
```
cbeighley: Doesn't look like time is used here? Were we going to automatically add a time suffix the to the filename?

denniszollo:
@cbeighley take a look at the first commit: https://github.com/denniszollo/piksi_tools/commit/3e36a66e7cb9be1bdf47358017f7546c3dfcc92a

cbeighley:
Cool, take out the unused import? then lgtm

denniszollo:
@cbeighley  but do you think it should be a prefix or the entire filename which is passed?  my vote is to go back to the prefix (first commit), in which case the time import is till relevant.

cbeighley:
Probably entire filename, it's more general.

PR #61, opened 2015-06-11
Adds Arch support

### LINKED ISSUE #45: setup.sh for other systems
On Arch Linux, `setup.sh` doesn't work because it relies on APT. Could a list of requirements as a .txt or less platform-dependent (e.g. checking `/etc/os-release`) `setup.sh` code be created?

tikiking1:
Also partially resolves #45.

### DIFF
--- a/piksi_tools/__init__.py
+++ b/piksi_tools/__init__.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python2
 # Copyright (C) 2011-2015 Swift Navigation Inc.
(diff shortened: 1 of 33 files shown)


henryhallam:
Maybe some apple-specific install-time thing that patches the shebangs... gross...

henryhallam:
@tikiking1 This looks good to me aside from the OSX incompatibility.  It would be nice if we can resolve that so we can merge it.  Perhaps leaving the shebang as 'python' but installing it under a python2-virtualenv on Arch would help? I'm afraid I'm pretty unfamiliar with Arch.

PR #66, opened 2015-06-17
[WIP] Bootloader tests

cbeighley:
For automated testing of the bootloader with a Jenkins job, using a card on Mufasa. Feedback appreciated @fnoble @mookerji @denniszollo. Currently doesn't have a good way to throw an error on a timeout, but I'm going to add this.


cbeighley:
retest this please

cbeighley:
Now wraps any operation that can hang in a timeout that will throw an Exception upon timing out.

piksi_tools/bootload.py
```diff
@@ -58,11 +58,13 @@ def __exit__(self, *args):
...
 
   def stop(self):
+    """ Remove Bootloader instance callbacks from serial link. """
     self.stopped = True
     self.link.remove_callback(self._deprecated_callback, SBP_MSG_BOOTLOADER_HANDSHAKE_DEPRECATED)
     self.link.remove_callback(self._handshake_callback, SBP_MSG_BOOTLOADER_HANDSHAKE_DEVICE)
 
   def _deprecated_callback(self, sbp_msg):
+    """ Bootloader handshake for deprecated message ID. """
     if len(sbp_msg.payload)==1 and struct.unpack('B', sbp_msg.payload[0])==0:
```
mookerji: Perhaps we want to use the generated code here?

piksi_tools/bootload.py
```diff
@@ -88,6 +105,10 @@ def wait_for_handshake(self, timeout=None):
     return True
 
   def reply_handshake(self):
+    """
+    Send a handshake to Piksi bootloader to request bootloader mode. Should be
+    sent after receiving handshake from Piksi bootloader.
+    """
     # < 0.45 of SBP protocol, reuse single handshake message.
     if self.sbp_version < (0, 45):
       self.link.send(SBP_MSG_BOOTLOADER_HANDSHAKE_DEPRECATED, '\x00')
```
mookerji: Same note about generated code from libsbp Python.

piksi_tools/bootload.py
```diff
@@ -95,6 +116,7 @@ def reply_handshake(self):
       self.link.send(SBP_MSG_BOOTLOADER_HANDSHAKE_HOST, '\x00')
 
   def jump_to_app(self):
+    """ Request Piksi bootloader jump to application. """
     self.link.send(SBP_MSG_BOOTLOADER_JUMP_TO_APP, '\x00')
```
mookerji: Here too.

tests/test_bootloader.py
```diff
@@ -0,0 +1,362 @@
...
+        link.start()
+        with Bootloader(link) as piksi_bootloader:
+          with Timeout(10) as timeout:
+            piksi_bootloader.wait_for_handshake()
+          with flash.Flash(link, flash_type='NAP', sbp_version=piksi_bootloader.version) \
+              as piksi_flash:
+            with Timeout(250) as timeout:
+              piksi_flash.write_ihx(NAP_FW)
+
+#  def test_program_btldr(self):
```
mookerji: Dead code => delete?

tests/test_bootloader.py
```diff
@@ -0,0 +1,362 @@
+#!/usr/bin/python
+
```
mookerji: nit: missing file copyright header

tests/test_bootloader.py
```diff
@@ -0,0 +1,362 @@
...
+# Firmware to flash in tests.
+STM_FW_URL = \
+  "http://downloads.swiftnav.com/piksi_v2.3.1/stm_fw/piksi_firmware_v0.17.hex"
+NAP_FW_URL = \
+  "http://downloads.swiftnav.com/piksi_v2.3.1/nap_fw/swift_nap_v0.13.hex"
+STM_FW = None
+NAP_FW = None
+
+class Heartbeat():
```
mookerji: nit: inherit from `object`.

Also, we may already have a Heartbeat callback. If not, this probably belongs in the main package for libsbp or piksi_tools somewhere.

tests/test_bootloader.py
```diff
@@ -0,0 +1,362 @@
...
+  """
+  Set Piksi into a known state (STM / NAP firmware).
+
+  Parameters
+  ==========
+  port : string
+    Filepath of virtual com port attached to Piksi.
+  """
+  if VERBOSE: print "Setting up Piksi at '%s' to known state" % port
+  with serial_link.get_driver(use_ftdi=False, port=port) as driver:
```
mookerji: Perhaps ftdi should be configurable (but default) as an argument?

tests/test_bootloader.py
```diff
@@ -0,0 +1,362 @@
...
+    return self
+
+  def __exit__(self, *args):
+    self.cancel()
+
+  def cancel(self):
+    """ Cancel scheduled Exception. """
+    signal.alarm(0)
+
+def set_known_state(port):
```
mfine: This is a super useful library function we should reuse elsewhere. Would be nice to lift this stuff out and make it more available and incorporate it in to our other testing tooling.

tests/test_bootloader.py
```diff
@@ -0,0 +1,362 @@
...
+    Filepath of virtual com port attached to Piksi.
+  """
+  if VERBOSE: print "Setting up Piksi at '%s' to known state" % port
+  with serial_link.get_driver(use_ftdi=False, port=port) as driver:
+    with Handler(driver.read, driver.write) as link:
+      link.start()
+
+      # Wait until we receive a heartbeat or bootloader handshake so we
+      # know what state Piksi is in.
+      if VERBOSE: print "  Waiting for Heartbeat or Bootloader Handshake"
```
mookerji: nit: newline

tests/test_bootloader.py
```diff
@@ -0,0 +1,362 @@
...
+
+        with flash.Flash(link, flash_type="M25",
+                 sbp_version=piksi_bootloader.sbp_version) as piksi_flash:
+          # Write NAP hexfile.
+          if VERBOSE: print "  Programming NAP"
+          with Timeout(250) as timeout:
+            piksi_flash.write_ihx(NAP_FW)
+
+        # Jump to the application firmware.
+        if VERBOSE: print "  Jumping to application"
```
mookerji: nit: newline

tests/test_bootloader.py
```diff
@@ -0,0 +1,362 @@
...
+        piksi_bootloader.jump_to_app()
+
+        if VERBOSE: print ""
```
mookerji: nit: newline

tests/test_bootloader.py
```diff
@@ -0,0 +1,362 @@
...
+
+class TestBootloader(unittest.TestCase):
+  """
+  Piksi bootloader tests.
+  """
+
+  @classmethod
+  def setUpClass(self):
```
mookerji: nit: camel case

tests/test_bootloader.py
```diff
@@ -0,0 +1,362 @@
...
+  parser.add_argument("-v", "--verbose",
+                      default=False, action="store_true",
+                      help="print more verbose output")
+  parser.add_argument('unittest_args', nargs='*')
+  return parser.parse_args()
+
+def main():
+  args = get_args()
+
+  global PORT1
```
mookerji: No globals.

tests/test_bootloader.py
```diff
@@ -0,0 +1,362 @@
...
+          with Timeout(5) as timeout:
+            while not heartbeat.received and not piksi_bootloader.handshake_received:
+              time.sleep(0.1)
+          if VERBOSE: print "  Received Heartbeat or Bootloader Handshake"
+          # If Piksi is in the application, reset it into the bootloader.
+          if heartbeat.received:
+            if VERBOSE: print "  Resetting Piksi"
+            link.send(SBP_MSG_RESET, "")
+
+      with Bootloader(link) as piksi_bootloader:
```
mfine: Do you need a new Bootloader context here? Can you reuse the previous bootloader? Think you're trying to test these out in isolation, but looking at the code there's not too much to Bootloader's context.

tests/test_bootloader.py
```diff
@@ -0,0 +1,362 @@
...
+
+# VCP to communicate with Piksi Under Test.
+PORT1 = None
+# VCP to communicate with second Piksi connected via UART to Piksi Under Test.
+PORT2 = None
+
+VERBOSE = False
+
+# Firmware to flash in tests.
+STM_FW_URL = \
```
mfine: maybe pull these values from ENV or arguments with sensible defaults?

tests/test_bootloader.py
```diff
@@ -0,0 +1,362 @@
...
+          # Write STM firmware.
+          if VERBOSE: print "  Programming STM"
+          with Timeout(100) as timeout:
+            piksi_flash.write_ihx(STM_FW, erase=False)
+
+        with flash.Flash(link, flash_type="M25",
+                 sbp_version=piksi_bootloader.sbp_version) as piksi_flash:
+          # Write NAP hexfile.
+          if VERBOSE: print "  Programming NAP"
+          with Timeout(250) as timeout:
```
mfine: Maybe add some constants for all these timeouts?

tests/test_bootloader.py
```diff
@@ -0,0 +1,362 @@
...
+        with Bootloader(link) as piksi_bootloader:
+          with Heartbeat(link) as heartbeat:
+            with Timeout(10) as timeout:
+              while not heartbeat.received and not piksi_bootloader.handshake_received:
+                time.sleep(0.1)
+            # If Piksi is in the application, reset it into the bootloader.
+            if heartbeat.received:
+              link.send(SBP_MSG_RESET, "")
+
+        with Bootloader(link) as piksi_bootloader:
```
mfine: Again, I think you could reuse that piksi_bootloader from above context.

tests/test_bootloader.py
```diff
@@ -0,0 +1,362 @@
...
+NAP_FW = None
+
+class Heartbeat():
+  """
+  Handle receiving heartbeat messages from Piksi. If we receive a heartbeat
+  from Piksi, we know that Piksi is in the application firmware.
+  """
+  def __init__(self, link):
+    self.received = False
+    self.link = link
```
mookerji: It's not obvious to me that you need context manager definition here, and I don't even think this should know anything about the serial connection. It would simplify things to do something like 

```python
class Hearbeat(object):
  def __init__(self):
    self.received = False
  
  def __call__(self):
    self.received = True


link.add_callback(Heartbeat, SBP_MSG_HEARTBEAT)
```

tests/test_bootloader.py
mookerji: +1

tests/test_bootloader.py
```diff
@@ -0,0 +1,362 @@
...
+  if VERBOSE: print "Setting up Piksi at '%s' to known state" % port
+  with serial_link.get_driver(use_ftdi=False, port=port) as driver:
+    with Handler(driver.read, driver.write) as link:
+      link.start()
+
+      # Wait until we receive a heartbeat or bootloader handshake so we
+      # know what state Piksi is in.
+      if VERBOSE: print "  Waiting for Heartbeat or Bootloader Handshake"
+      with Bootloader(link) as piksi_bootloader:
+        with Heartbeat(link) as heartbeat:
```
mookerji: Yeah, if you removed the context management stuff and just used vanilla callbacks, then this following bit will be simpler to follow.

tests/test_bootloader.py
```diff
@@ -0,0 +1,362 @@
...
+
+      with Bootloader(link) as piksi_bootloader:
+        # Set Piksi into bootloader mode.
+        with Timeout(10) as timeout:
+          piksi_bootloader.wait_for_handshake()
+        piksi_bootloader.reply_handshake()
+        if VERBOSE: print "  Received bootloader handshake"
+
+        with flash.Flash(link, flash_type="STM",
+                 sbp_version=piksi_bootloader.sbp_version) as piksi_flash:
```
mookerji: nit: indent

tests/test_bootloader.py
mookerji: +1

tests/test_bootloader.py
```diff
@@ -0,0 +1,362 @@
...
+          with Timeout(250) as timeout:
+            piksi_flash.write_ihx(NAP_FW)
+
+        # Jump to the application firmware.
+        if VERBOSE: print "  Jumping to application"
+        piksi_bootloader.jump_to_app()
+
+        if VERBOSE: print ""
+
+class TestBootloader(unittest.TestCase):
```
mookerji: random suggestion: I don't know how these tests are invoked/discovered, but I've had some good experiences with pytest and pytest's setup fixtures (https://pytest.org/latest/example/simple.html) for writing different integration test cases.

piksi_tools/bootload.py
cbeighley: @mookerji I think this will only work for the most recent version of the message at the moment, as there are fields in the older version that aren't present in the message from the device with the old bootloader.

```
  File "/Users/colin/miniconda/lib/python2.7/site-packages/sbp-0.45.post8_g805031f-py2.7.egg/sbp/client/handler.py", line 140, in run
    self.call(msg)
  File "/Users/colin/miniconda/lib/python2.7/site-packages/sbp-0.45.post8_g805031f-py2.7.egg/sbp/client/handler.py", line 217, in call
    callback(msg)
  File "./bootload.py", line 68, in _deprecated_callback
    bh = MsgBootloaderHandshakeDevice(sbp_msg)
  File "/Users/colin/miniconda/lib/python2.7/site-packages/sbp-0.45.post8_g805031f-py2.7.egg/sbp/bootload.py", line 97, in __init__
    self.from_binary(sbp.payload)
  File "/Users/colin/miniconda/lib/python2.7/site-packages/sbp-0.45.post8_g805031f-py2.7.egg/sbp/bootload.py", line 113, in from_binary
    p = MsgBootloaderHandshakeDevice._parser.parse(d)
  File "/usr/local/lib/python2.7/site-packages/construct/core.py", line 188, in parse
    return self.parse_stream(BytesIO(data))
  File "/usr/local/lib/python2.7/site-packages/construct/core.py", line 198, in parse_stream
    return self._parse(stream, Container())
  File "/usr/local/lib/python2.7/site-packages/construct/core.py", line 670, in _parse
    subobj = sc._parse(stream, context)
  File "/usr/local/lib/python2.7/site-packages/construct/core.py", line 365, in _parse
    raise FieldError(sys.exc_info()[1])
FieldError: expected 4, found 1
```

We could add a message class for the old message?

piksi_tools/bootload.py
cbeighley: @mookerji Is there an example for how to use the message class to send messages, rather than decoding a received message in a callback?

tests/test_bootloader.py
cbeighley: Done, uncommented out unimplemented tests.

tests/test_bootloader.py
cbeighley: Fixed

tests/test_bootloader.py
cbeighley: Changed to inherit `object`. There isn't a general purpose `Heartbeat` class defined anywhere at the moment.

cbeighley: @mookerji Where would be an appropriate place to put a `Heartbeat` class? `libsbp/python/sbp/client`?

cbeighley: @mookerji Put `heartbeat.py` in `piksi_tools/piksi_tools`. `serial_link.py` and `console/system_monitor_view.py` still have callbacks registered directly to `SBP_MSG_HEARTBEAT` - seemed clearest this way.

tests/test_bootloader.py
cbeighley: Factored this and Timeout out into piksi_tools/utils.py

tests/test_bootloader.py
cbeighley: Fixed

tests/test_bootloader.py
cbeighley: This is a method inherited from unittest.TestCase so unfortunately I can't do much about it.

tests/test_bootloader.py
cbeighley: Fixed

tests/test_bootloader.py
cbeighley: I thought it was nice to have them just versioned in the source - they should stay as constant as possible for repeatability.

tests/test_bootloader.py
cbeighley: Is there a way to pass arguments into a unittest.TestCase class? This was the only way I could think of to do it.

tests/test_bootloader.py
cbeighley: I think it should as the heartbeat message now sends the versioning info which needs to be parsed.

piksi_tools/bootload.py
mookerji: https://github.com/swift-nav/libsbp/blob/master/python/sbp/deprecated.py?

piksi_tools/bootload.py
mookerji: You should be able to construct an object inline and send that over. 

Construction example for another message:
https://github.com/swift-nav/sbp_log_analysis/blob/master/tests/test_rtk_fix_loss.py#L29

From, https://github.com/swift-nav/libsbp/blob/master/python/sbp/deprecated.py#L28,
```
MsgBootloaderHandshakeDeprecated(0x1337, 0).pack 
```
should work.

tests/test_bootloader.py
mookerji: God damnit, Python.

tests/test_bootloader.py
mookerji: You should be able to define  `def __init__(self, port, verbose, etc)` for `test_case = TestBootloader()` and call its superclass constructor as part of that.

piksi_tools/heartbeat.py
```diff
@@ -0,0 +1,41 @@
...
+# This source is subject to the license found in the file 'LICENSE' which must
+# be be distributed together with this source. All other rights reserved.
+#
+# THIS CODE AND INFORMATION IS PROVIDED "AS IS" WITHOUT WARRANTY OF ANY KIND,
+# EITHER EXPRESSED OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE IMPLIED
+# WARRANTIES OF MERCHANTABILITY AND/OR FITNESS FOR A PARTICULAR PURPOSE.
+
+from sbp.system import SBP_MSG_HEARTBEAT, MsgHeartbeat
+
+class Heartbeat(object):
```
mookerji: I still think the link and callback stuff should be left out of the Heartbeat thing (see previous comment, which has since disappeared). We have an existing pattern of simple callbacks managed by at a high-level by the app (https://github.com/swift-nav/piksi_tools/blob/4d4dc570433ecaf190b6c4631121cccb2ef33d46/piksi_tools/serial_link.py#L171) that should probably be followed here. If it's not doing any resource management, it probably doesn't need context management either (which should simplify the code using it below).

tests/test_bootloader.py
```diff
@@ -0,0 +1,269 @@
...
+    pass
+
+  """
+  Test if two Piksies can set eachother into bootloader mode (should fail).
+  """
+  def test_two_piksies_btldr_mode(self):
+    if PORT2 is None:
+      return
+
+  """ Test if two Piksies can simultaneously bootload. """
```
mookerji: What's going on with these docstrings up here? Also, if you're planning on filling in these tests later, perhaps add a TODO or `raise NotImplementedError("TODO")`

piksi_tools/heartbeat.py
cbeighley: I agree on the link side, but I'm not sure how the context management line gets drawn between something like Heartbeat and Flash - the __exit__  function does the same thing in either case, removes the callbacks from the handler.

cbeighley: For something like this bootloader testing stuff, you need to instantiate a lot of `Heartbeat` classes for which the callbacks won't be removed unless you specifically remember to do so.

cbeighley: Although I guess the link is required to be passed for the context management.

tests/test_bootloader.py
cbeighley: Fixed

tests/test_bootloader.py
cbeighley: Doesn't seem to be that simple. `unittest` does a bunch of magic behind the scenes and seems to usually expect to be passed the class rather than instances of it. Been playing around with this for an hour and can't figure out how to do it.

mookerji: Ironic that a unit testing library would require that passing global variables as a means of getting configuration state.

mookerji: http://stackoverflow.com/questions/1842168/python-unit-test-pass-command-line-arguments-to-setup-of-unittest-testcase?

Oh man, this is like the first example with py.test: https://pytest.org/latest/example/simple.html

cbeighley: I think you're not really supposed to passing stuff in - it's meant to implement unit tests rather than integration tests. I'll give that post a shot.

piksi_tools/bootload.py
```diff
@@ -58,25 +58,44 @@ def __exit__(self, *args):
...
-    if len(sbp_msg.payload)==1 and struct.unpack('B', sbp_msg.payload[0])==0:
+    """ Bootloader handshake for deprecated message ID. """
+    hs_device = MsgBootloaderHandshakeDeprecated(sbp_msg)
+    if len(hs_device.handshake)==1 and hs_device.handshake[0]==0:
       # == v0.1 of the bootloader, returns hardcoded version number 0.
       self.version = "v0.1"
     else:
       # > v0.1 of the bootloader, returns git commit string.
-      self.version = sbp_msg.payload[:]
+      self.version = ''.join([chr(i) for i in hs_device.handshake])
```
mookerji: If this `hs_device.handshake` ends up being empty for some reason, will `self.version = ''`. Is that ok?

cbeighley: Maybe I should make that a special case:

```
self.version = ''.join([chr(i) for i in hs_device.handshake])
if self.version = '':
  self.version = "Unknown"
```

piksi_tools/stm_unique_id.py
```diff
@@ -14,36 +14,51 @@
...
+      link to register Heartbeat message callback with
+    """
     self.unique_id_returned = False
-    # SBP version is unset in older devices.
-    self.sbp_version = (0, 0)
     self.unique_id = None
     self.link = link
-    link.add_callback(self.receive_heartbeat, SBP_MSG_HEARTBEAT)
+    self.heartbeat = Heartbeat(link)
     link.add_callback(self.receive_stm_unique_id_callback, SBP_MSG_STM_UNIQUE_ID_DEVICE)
```
mookerji: I'm still failing to see how this heartbeat here is different than the stm callback you have on the next line. You can construct a Hearbeat object implementing `__call__` and have equivalent behavior without having this heartbeat callback be aware of the serial connection. Yeah, you have to remove the callback, but you can just do that like how you're removing the stm callback at 44.

cbeighley: Yeah, this is a case where you wouldn't use the context handling as the Heartbeat object has to persist in between STMUniqueID calls, and so you need to clean up upon exit (this should get added here). What I'm more concerned about are scripts where you may be constantly creating/destroying the Heartbeats. I'll change it though.

tests/test_bootloader.py
cbeighley: For some reason this is breaking the skipIf for Travis. I might try using pytest.

piksi_tools/stm_unique_id.py
cbeighley: Done.

piksi_tools/bootload.py
cbeighley: Done.

tests/test_bootloader.py
cbeighley: Gonna punt on using py.test for the moment and keep the globals.

mookerji:
Good for this to get merged whenever @cbeighley thinks its ready.

cbeighley:
Completion criteria: HITL job that runs latest bootloader release / piksi_tools/master through suite of bootloader tests when piksi_tools/master is updated. Would like to extend this to running suite of tests over all bootloaders / all firmware releases at some point but I'm going to call that out of scope for this sprint.

PR #67, opened 2015-06-17
Handler starts the link

mfine:
Handler takes care of starting the link. See https://github.com/swift-nav/libsbp/pull/170. Needs to go in first and pick up a new sbp dependency.

/cc @cbeighley

### DIFF
--- a/piksi_tools/acq_results.py
+++ b/piksi_tools/acq_results.py
@@ -107,7 +107,6 @@ def main():
     with Handler(driver.read, driver.write) as link:
       link.add_callback(serial_link.printer, SBP_MSG_PRINT)
       acq_results = AcqResults(link)
-      link.start()
 
       try:
         while True:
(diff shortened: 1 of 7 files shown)


cbeighley:
lgtm

denniszollo:
sbp log analysis device runner will need to be udpdated as well

denniszollo:
Is there a possibility that early messages are ignored because the link starts before callbacks are registered?

cbeighley:
I think you'll drop those messages anyway.

mfine:
Is there anywhere that we're depending on receiving messages before callbacks are registered? That sounds like a race condition we should avoid.

mfine:
@denniszollo sbp_log_analysis isn't using the context manager, shouldn't need any updates to take this change.

mfine:
When the `libsbp` change gets ok'd, I'll bring this in with a new `libsbp` version.

mfine:
Bringing this in.

PR #72, opened 2015-06-18
Fix message handling

mfine:
Man, these weren't working at all. Ran them against an old device and look ok, but should be thoroughly vetted on new bootloader, new firmware.

/cc @cbeighley


cbeighley:
Hehe yeah I actually fixed this in my bootloader-testing branch too. lgtm

PR #71, opened 2015-06-18
Generate settings as JSON file

mfine:
Simple utility to generate our settings as a JSON file for consumption by other tooling.

Generated JSON looks like:

```json
{
  "ext_events": {
    "edge_trigger": "None"
  },
  "float_kf": {
    "amb_init_var": "1e+25",
    "code_var": "40000",
    "new_amb_var": "1e+25",
    "phase_var": "0.0144"
  },
  "frontend": {
    "antenna_selection": "External"
  },
  "iar": {
    "code_var": "40000",
    "phase_var": "0.0144"
  },
  "sbp": {
    "obs_msg_max_size": "104"
  },
  "simulator": {
    "base_ecef_x": "-2700303.10144",
    "base_ecef_y": "-4292474.39651",
    "base_ecef_z": "3855434.34087",
    "cn0_sigma": "0.300000011921",
    "enabled": "False",
    "mode_mask": "15",
    "num_sats": "9",
    "phase_sigma": "0.0299999993294",
    "pos_sigma": "1.5",
    "pseudorange_sigma": "4",
    "radius": "100",
    "speed": "4",
    "speed_sigma": "0.15000000596"
  },
  "solution": {
    "dgnss_filter": "Fixed",
    "dgnss_solution_mode": "Low Latency",
    "known_baseline_d": "0",
    "known_baseline_e": "0",
    "known_baseline_n": "0",
    "output_every_n_obs": "2",
    "soln_freq": "10"
  },
  "surveyed_position": {
    "broadcast": "False",
    "surveyed_alt": "0",
    "surveyed_lat": "0",
    "surveyed_lon": "0"
  },
  "system_info": {
    "firmware_built": "Jun 16 2015 00:15:21",
    "firmware_version": "v0.17",
    "hw_revision": "piksi_2.3.1",
    "nap_channels": "11",
    "nap_fft_index_bits": "13",
    "nap_version": "v0.13",
    "serial_number": "36828"
  },
  "system_monitor": {
    "heartbeat_period_milliseconds": "1000",
    "watchdog": "True"
  },
  "telemetry_radio": {
    "configuration_string": "AT&F,ATS1=115,ATS2=128,ATS5=0,AT&W,ATZ"
  },
  "track": {
    "iq_output_mask": "0"
  },
  "uart_ftdi": {
    "baudrate": "1000000",
    "mode": "SBP",
    "sbp_message_mask": "65535"
  },
  "uart_uarta": {
    "baudrate": "115200",
[...]

PR #73, opened 2015-06-18
Versions Discovery

mfine:
Determine the versions of the bootloader and sbp and write them to a file. Contents of file look like:

```json
{
  "bootloader": "v1.2",
  "sbp": 0
}
```

Are there more versions to capture from the device?

/cc @mookerji @cbeighley

### DIFF
--- a/piksi_tools/versions.py
+++ b/piksi_tools/versions.py
@@ -48,7 +48,7 @@ class Versions(object):
     self.handshake_received = True
 
   def _handshake_callback(self, sbp_msg):
-    self.versions['bootloader'] = MsgBootloaderHandshkeDevice(sbp_msg).version
+    self.versions['bootloader'] = MsgBootloaderHandshakeDevice(sbp_msg).version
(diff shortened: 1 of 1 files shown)


mookerji:
From device1_settings.txt:

```
nap_version          v0.12-39-gbd5edfc
serial_number        1226
firmware_version     v0.16-baseline_update-10-g44115b9
hw_revision          piksi_2.3.1
```
 ... in addition to what you mentioned.

cbeighley:
There are also the NAP key, STM Unique ID, and NAP device dna, but they're not versions persay.

mfine:
Those should be covered in settings - see the sample output in #71 under `system_info`.

But hmm, maybe we should just collect all the things into one output? Including the other stuff like NAP key, STM UniqueID, and NAP device DNA?

Instead of settings - it'd be more like diagnostics or something?

mookerji:
:+1: single data file

mfine:
Closing this in favor of merging it into #71.

PR #74, opened 2015-06-19
New mavlink bridge

denniszollo:
This PR groups the python console, system monitor, and new SBP Relay tab into an "Advanced" tab
It also adds a semi-generic ability to forward SBP messages over UDP either at the command line or whilst running the console.  

The forwarding ability is for use with 3dr Pixhawk integration


denniszollo:
@mfine, @fnoble, @mookerji.  Anytime for code review of this?

denniszollo:
PYTHONPATH=
screen shot 2015-06-18 at 6 34 36 pm

piksi_tools/console/sbp_relay_view.py
```diff
@@ -0,0 +1,117 @@
...
+# Contact: Colin Beighley <colin@swift-nav.com>
+#
+# This source is subject to the license found in the file 'LICENSE' which must
+# be be distributed together with this source. All other rights reserved.
+#
+# THIS CODE AND INFORMATION IS PROVIDED "AS IS" WITHOUT WARRANTY OF ANY KIND,
+# EITHER EXPRESSED OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE IMPLIED
+# WARRANTIES OF MERCHANTABILITY AND/OR FITNESS FOR A PARTICULAR PURPOSE.
+
+from traits.api import HasTraits, String, Button, Instance, Int, Bool, \
```
mookerji: Cleaner to do a qualified `import trails.api as api` or something.

mookerji: Cleaner to do a qualified `import trails.api as api` or something.

piksi_tools/console/sbp_relay_view.py
```diff
@@ -0,0 +1,117 @@
...
+                       on_trait_change, Enum, on_trait_change
+from traitsui.api import View, Item, VGroup, UItem, HGroup, TextEditor, spring
+from piksi_tools.udp_bridge import open_socket, register_udp_callbacks, \
```
mookerji: ditto.

piksi_tools/udp_bridge.py
```diff
@@ -0,0 +1,93 @@
...
+"""
+
+from sbp.client.drivers.pyserial_driver import PySerialDriver
+from sbp.client.handler import Handler
+from sbp.observation import SBP_MSG_OBS, SBP_MSG_BASE_POS
+
+import socket
+import struct
+import time
+
```
mookerji: An integration test for this stuff would be cool. I.e., some way of reading a message log, serializing, sending it over UDP and then catching it on the other side. 

Here's an example of how to setup a mock socket server: https://github.com/swift-nav/pySTR4500/blob/master/test/test_client.py. 

You can also construct objects inline, `.pack()`, and dispatch over UDP. For inline construction, see: https://github.com/swift-nav/sbp_log_analysis/blob/master/tests/test_rtk_fix_loss.py#L29

piksi_tools/udp_bridge.py
```diff
@@ -0,0 +1,93 @@
...
+DEFAULT_UDP_ADDRESS = "127.0.0.1"
+DEFAULT_UDP_PORT    = 13320
+
+def get_args():
+  """
+  Get and parse arguments.
+  """
+  import argparse
+  parser = argparse.ArgumentParser(description="Swift Navigation SBP Example.")
```
mookerji: Update help message.

piksi_tools/udp_bridge.py
```diff
@@ -0,0 +1,93 @@
...
+def unregister_udp_callbacks(link, func_list, msg_list):
+  for eachfunc,eachmsg in zip(func_list, msg_list):
+    link.remove_callback(eachfunc, eachmsg)
+
+def main():
+  args = get_args()
+
+  with PySerialDriver(args.serial_port[0], args.baud[0]) as driver:
+    with Handler(driver.read, driver.write) as handler:
+      udp = open_socket()
```
mookerji: want to have some more robust socket handling. crib: https://docs.python.org/2/library/socket.html#example

piksi_tools/console/sbp_relay_view.py
```diff
@@ -0,0 +1,117 @@
...
+    self.python_console_cmds = {
+      'update': self
+    }
+
+  def update_msgs(self):
+    """
+    Updates the instance variable msgs which store the msgs that we will send over UDP
+    """
+    if self.msg_enum =='Observations':
+      self.msgs=OBS_MSGS
```
mookerji: nit: spacing with `=`.

Depending on what text editor you're using, you should be able to find a pep8/pyflakes thing that will link formatting things.

piksi_tools/udp_bridge.py
```diff
@@ -0,0 +1,93 @@
...
+                      default=[DEFAULT_UDP_ADDRESS], nargs=1,
+                      help="specify the UDP IP Address to use.")
+  parser.add_argument("-p", "--udp-port",
+                      default=[DEFAULT_UDP_PORT], nargs=1,
+                      help="specify the UDP Port to use.")
+  return parser.parse_args()
+
+def open_socket():
+  return socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
+
```
mookerji: I was going to suggest refactoring this, but then I realized that this really ought to exist as a TCP/UDP network driver in the libsbp client module. This then would become the CLI front-end for using it (just like serial_link).

See: 
1. https://github.com/swift-nav/libsbp/blob/master/python/sbp/client/handler.py
2. https://github.com/swift-nav/libsbp/tree/master/python/sbp/client/drivers

mookerji:
I would probably talk to @mfine about refactoring the UDP stuff here to be a lightweight networking handler in the libsbp client.

mfine:
Yeah, makes sense. Would want potentially a UDP driver (not this use case), and a UDP logger (this use case).

piksi_tools/console/sbp_relay_view.py
fnoble: @mookerji : possibly but thats not the Traits convention, better to be consistent with the rest of the code and typical library usage

denniszollo:
In the short term, I think it belongs in PIksi_Tools.  In the future, I agree we could lift the udp stuff into LibSBP. My takeaway from the feedback are the following action items:
Fix nits
Be a little more robust in socket opening as per the python docs
Add a simple unit test (have travis run it)

I don't think LibSBP should have any CLI's except the examples.

piksi_tools/console/sbp_relay_view.py
fnoble: maybe a qualified import would be better here for the `udp_bridge` import.

piksi_tools/udp_bridge.py
fnoble: @mookerji : I'd suggest leaving this here for now and it can graduate to libsbp in time as it proves to be more widely useful. I think you made the suggestion to move it to piksi_tools when this code was originally submitted as a PR to libsbp :)

https://github.com/swift-nav/libsbp/pull/141

fnoble:
@denniszollo : lgtm, if you address the points you mentioned I'm happy to merge it!

I think people in the community are going to be very happy about this!

piksi_tools/udp_bridge.py
denniszollo: The one thing that is hard from a CLI perspective is exactly how to filter messages.  What if you only want to broadcast 10 SBP Messages over UDP?  What if you want to filter the rate of some of the messages broadcast? I think it would be pretty hard to make a full featured command line tool and we might be better off just providing examples for people to do there own python.

fnoble: Yeah, I'd either just let everything through for now or at most do like the console and have a simple flag to restrict to a preset set of messages.

mookerji: Well, I meant that the tool should be here, and that was when it seemed Mavlink-specific. Now it seems like a very generic UDP driver/logger that supports SBP message callback. This follows a very similar pattern to serial link: the command line tool is here and it's used in the console, but the underlying behavior is in the library. 

Regarding other things like filtering, backpressure, etc.: that's the kind of behavior that should be designed/implemented into a library component (note: definitely not saying we have to do this now).

I suppose I'm fine letting this go in as well, but it's also strikes me as classic technical debt that could be avoided with not that much work.

mfine: :+1: Let's lift the platform-oriented bits into the sbp library, and keep piksi_tools and related artifacts more product-oriented. Then it's a lot easier to reuse and compose the platform-oriented bits for other product efforts. This `udp_bridge` is tying up two very useful primitives into a single non-reuseable tool: forwarding SBP messages as UDP datagrams and registering a callback for multiple message types. Anyone looking to take advantage of either of these primitives would have to cut-and-paste the code here (as this code was cut-and-pasted from libsbp's examples). It should be pretty straightforward to factor both of these ideas out into libsbp and make them generally available for everyone. Here's my quick pass at both:

UdpLogger: https://github.com/swift-nav/libsbp/pull/173
Callbacks for multiple messages: https://github.com/swift-nav/libsbp/pull/174

What do you think?

piksi_tools/udp_bridge.py
mfine: Do we need more robust socket handling with UDP? Most of that sheet is about handling TCP issues, and UDP is kind of super simple (fire and forget). Think it should be ok to leave the socket error handling to a minimum. What do you think?

PR #82, opened 2015-06-23
New settings

mfine:
Support the new settings and the old settings. This is pretty terrible. Better approaches to this welcome.

/cc @denniszollo @mookerji @fnoble


piksi_tools/bootload.py
```diff
@@ -72,9 +72,9 @@ def _deprecated_callback(self, sbp_msg):
     self.handshake_received = True
 
   def _handshake_callback(self, sbp_msg):
-    self.version = sbp_msg.version
-    self.sbp_version = ((MsgBootloaderHandshakeDevice(sbp_msg).flags >> 8) & 0xF,
-                        MsgBootloaderHandshakeDevice(sbp_msg).flags & 0xF)
+    msg = MsgBootloaderHandshakeDevice(sbp_msg)
+    self.version = msg.version
+    self.sbp_version = ((msg.flags >> 8) & 0xFF, msg.flags & 0xFF)
```
cbeighley: The packing here for the SBP version is different than the heartbeat flags packing, which is a little confusing. Probably doesn't matter though.

mfine: Yeah, unfortunate - would have like to be similar, but the low and high bits of the heartbeat status field were spoken for.

cbeighley:
Just to be clear, can you explain what the old sequence of messages to read/write settings to the hardware was and what it will be now with this change?

mfine:
### Old way:

Host sends `SBP_MSG_SETTINGS_READ_BY_INDEX` with the indexed setting to read (starts at 0).
Device sends back `SBP_MSG_SETTINGS_READ_BY_INDEX` with the setting value. When there is no value to read at an index, the device sends back a `SBP_MSG_SETTINGS_READ_BY_INDEX` with null payload.

Host sends `SBP_MSG_SETTINGS` with the settings value to write. Device sends back `SBP_MSG_SETTINGS` with the value written (this is not consumed in the console, and a hack is used to prevent an infinite loop :(

### New way:

Host sends `SBP_MSG_SETTINGS_READ_BY_INDEX_REQUEST` with the indexed setting to read (starts at 0).
Device sends back `SBP_MSG_SETTINGS_READ_BY_INDEX_RESPONSE` with the setting value. When there is no value to read at an index, the device sends back `SBP_MSG_SETTINGS_READ_BY_INDEX_DONE`.

Host sends `SBP_MSG_SETTINGS_WRITE` with the settings value to write. Device does not send anything back.

mfine:
Rebased and all set to go.

PR #88, opened 2015-06-25
Improve bootloader, update log output

mfine:
Make the bootloader more robust to missing handshakes.

/cc @mookerji @cbeighley

### DIFF
--- a/piksi_tools/bootload.py
+++ b/piksi_tools/bootload.py
@@ -81,11 +81,16 @@ class Bootloader():
     if timeout is not None:
       t0 = time.time()
     self.handshake_received = False
+    expire = time.time() + 15.0
+    self.link.send(SBP_MSG_RESET, "")
     while not self.handshake_received:
       time.sleep(0.1)
       if timeout is not None:
         if time.time()-timeout > t0:
           return False
+      if time.time() > expire:
+        expire = time.time() + 15.0
+        self.link.send(SBP_MSG_RESET, "")
     return True
 
   def reply_handshake(self):
@@ -151,8 +156,6 @@ def main():
   with serial_link.get_driver(use_ftdi, port, baud) as driver:
     # Handler with context
     with Handler(driver.read, driver.write) as link:
-      link.send(SBP_MSG_RESET, "")
-      time.sleep(0.2)
       link.add_callback(serial_link.printer, SBP_MSG_PRINT)
 
       # Tell Bootloader we want to write to the flash.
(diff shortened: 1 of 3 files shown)


mookerji:
lgtm. @cbeighley ?

piksi_tools/bootload.py
```diff
@@ -81,11 +81,16 @@ def wait_for_handshake(self, timeout=None):
...
+    expire = time.time() + 15.0
+    self.link.send(SBP_MSG_RESET, "")
     while not self.handshake_received:
       time.sleep(0.1)
       if timeout is not None:
         if time.time()-timeout > t0:
           return False
+      if time.time() > expire:
+        expire = time.time() + 15.0
+        self.link.send(SBP_MSG_RESET, "")
```
cbeighley: I feel like it may be worth changing the name of this function entirely if we want to do this - seems like its more ```set_bootloader_mode```.

mfine: Right now it kind of lines up with `reply_handshake` - but maybe we should roll that send in as well and rename the function?

cbeighley: Or adding another function named that.

cbeighley: I've added a function for that in my bootloader-tests PR (https://github.com/swift-nav/piksi_tools/pull/66) but it probably makes more sense for it to be here.

mfine: I moved it to just `handshake()` - e.g., `bootloader.handshake()`. What do you think?

cbeighley: Yeah I think we should roll `reply_handshake` into it and have one function that takes the device from whatever state it's currently in and puts it in the bootloader, ready to flash. I think `handshake` would be good - do you think that its obvious that would take you from the application firmware into the bootloader?

mfine: Makes sense to me - I updated the patch. `bootloader.handshake()` seems like a good description - handshake as a verb.

piksi_tools/bootload.py
```diff
@@ -77,23 +77,26 @@ def _handshake_callback(self, sbp_msg):
                         MsgBootloaderHandshakeDevice(sbp_msg).flags & 0xF)
     self.handshake_received = True
 
-  def wait_for_handshake(self, timeout=None):
+  def handshake(self, timeout=None):
     if timeout is not None:
       t0 = time.time()
     self.handshake_received = False
+    expire = time.time() + 15.0
```
cbeighley: You may consider cherry picking piksi_tools.timeout.Timeout out of my bootlooader-tests PR and using that here.

cbeighley: Err, nevermind. Currently won't work on windows. Axe that idea.

mfine: Would be great to get a better pattern for timeouts - looking forward to it!

cbeighley:
lgtm

PR #95, opened 2015-06-29
Added option to save/load config from file

ungureanuvladvictor:
Some change is happening with the settings/fileio so we can keep this hanging until those get merged and I will update the PR.

This PR depends on https://github.com/swift-nav/libsbp/pull/188

/cc @denniszollo @mfine

### DIFF
--- a/piksi_tools/console/settings_view.py
+++ b/piksi_tools/console/settings_view.py
@@ -99,6 +99,8 @@ class Setting(SettingBase):
         new is not Undefined):
       if type(self.value) == unicode:
         self.value = self.value.encode('ascii', 'replace')
+      if self.name == "baudrate":
+        print(str(self.name).upper() + " FOR " + str(self.section).upper() + " CHANGED TO " + new)
       self.settings.set(self.section, self.name, self.value)
 
 class EnumSetting(Setting):


piksi_tools/console/settings_view.py
```diff
@@ -212,6 +227,40 @@ def _factory_default_button_fired(self):
...
+      blob = fio.read("config")
+      config_file.write(blob)
+
+  def load_settings_from_file(self, filepath):
+    fio = FileIO(self.link)
+    with open(filepath, 'r') as config_file:
+      blob = config_file.read()
+      fio.write("config", blob)
+    self.link.send(SBP_MSG_RESET, '')
+    self._settings_read_button_fired()
```
cbeighley: Caling `_settings_read_button_fired()` happens a couple other places too, but is kind of janky - can we refactor so other code that wants to trigger a settings read doesn't need to use the GUI hooks?

piksi_tools/console/settings_view.py
```diff
@@ -212,6 +227,40 @@ def _factory_default_button_fired(self):
...
+      filepath = os.path.join(dialog.directory, dialog.filename)
+      self.save_settings_to_file(filepath)
+    else:
+      print 'Unable to save file'
+
+  def save_settings_to_file(self, filepath):
+    fio = FileIO(self.link)
+    with open(filepath, 'w') as config_file:
+      blob = fio.read("config")
+      config_file.write(blob)
```
cbeighley: Is it going to be obvious that you're savings the current Piksi settings to a file, not the current set of settings in the console (that may or may not have been saved to Piksi's flash)?

piksi_tools/console/settings_view.py
```diff
@@ -212,6 +227,40 @@ def _factory_default_button_fired(self):
                         + "Are you sure you want to reset to factory defaults?"
     confirm_prompt.run(block=False)
 
+  def _settings_save_to_file_fired(self):
```
cbeighley: Is it going to be obvious that you're savings the current Piksi settings to a file, not the current set of settings in the console (that may or may not have been saved to Piksi's flash)?

cbeighley:
Maybe outside the scope of this PR, but it would be good to have some indicator of whether the current settings in the console have been saved to the Piksi flash - maybe change the color of the Save Settings to Flash button when settings in the console are changed and change it back to default when they are saved to the Piksi flash.

piksi_tools/console/settings_view.py
ungureanuvladvictor: I think we should change the logic. When a setting is changed in the UI it should be saved to flash also at that point. So we do not have any differences between UI/Flash settings. It was a bit confusing for me at the start what happens and I think it is the same for a random user who does not know the code.

fnoble: The problem is if you change a setting that makes it no longer possible to communicate with the device easily (such as changing baud rate), then a reset will bring it back.

fnoble: I guess they can always "reset to defaults" or worst case do a full bootload. @denniszollo @cbeighley , what do you guys think?

cbeighley: Could we add something to the settings.yaml to make the baud rate non-configurable for the FTDI USART through the console by default? I feel like there's more value in making this difficult to modify and avoiding people bricking their device than there is in keeping it configurable through the console. We'll have to provide some documentation around modifying the settings through SBP messages directly at some point anyways.

fnoble: I've seen situations where people needed to change this (and I definitely have several times myself). I kinda feel like it would be more confusing if the option was hidden. People do use the USB port for connection to systems other than just the console it seems.

ungureanuvladvictor: We should leave this settings to be accessible for the user, it is more annoying if you cannot do something related to your device.  I'll remove both save to flash and refresh and make the UI be in sync with flash at every change and then we can see how it looks.

cbeighley: Cool. Maybe we can print a very loud, obvious message when the baud rate is changed on any of the USARTs?

fnoble: Or just specify a lower bound of 9600 or something in the firmware that will still kinda work

PR #100, opened 2015-06-30
Verbose failures from check_device_details.

mookerji:
/cc @denniszollo


denniszollo:
Looks good, might you update the assert fail messages as well?

PR #104, opened 2015-07-02
CONSOLE: hide expert settings by default

henryhallam:
cc @denniszollo @gsmcmullin @fnoble  
Any setting category beginning with "expert" will be hidden unless -e is passed on the console command line.

### DIFF
--- a/piksi_tools/console/console.py
+++ b/piksi_tools/console/console.py
@@ -58,6 +58,8 @@ def get_args():
                       action="store_true")
   parser.add_argument('-t', '--toolkit', nargs=1, default=[None],
                       help="specify the TraitsUI toolkit to use, either 'wx' or 'qt4'.")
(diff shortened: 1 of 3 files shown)


denniszollo:
What do we think about adding an expert flag to the Yaml file instead?

denniszollo:
There are many settings that aren't in an "expert" group that are really expert settings, such as the kalman filter tuning parameters

denniszollo:
Also, what if the some settings in a group are expert and some are not?  I don't want to rename settings unless we have to.

cbeighley:
+1 I think a flag is the way to go

fnoble:
+1 on flag

henryhallam:
Even though it's not yet July 4th:
*waves flag*

How's this look?

denniszollo:
LGTM

PR #106, opened 2015-07-06
CONSOLE: visually distinguish fixed from float baselines

### LINKED ISSUE #101: Console: plot fixed baseline solutions in different color than float
or other means of distinguishing them.
@denniszollo how much work do you think this would be?

henryhallam:
Closes #101, sorta
@denniszollo @fnoble this look ok to you?

### DIFF
--- a/piksi_tools/console/baseline_view.py
+++ b/piksi_tools/console/baseline_view.py
@@ -46,6 +46,7 @@ class BaselineView(HasTraits):
   ns = List()
   es = List()
   ds = List()
+  fixeds = List()
 
   table = List()
 
@@ -93,6 +94,7 @@ class BaselineView(HasTraits):
           Item('reset_button', show_label=False),
           Item('reset_iar_button', show_label=False),
           Item('init_base_button', show_label=False),
+            
         ),
         Item(
           'plot',
@@ -128,9 +130,13 @@ class BaselineView(HasTraits):
     self.ns = []
     self.es = []
     self.ds = []
-    self.plot_data.set_data('n', [])
-    self.plot_data.set_data('e', [])
-    self.plot_data.set_data('d', [])
+    self.fixeds = []
+    self.plot_data.set_data('n_fixed', [])
+    self.plot_data.set_data('e_fixed', [])
+    self.plot_data.set_data('d_fixed', [])
+    self.plot_data.set_data('n_float', [])
+    self.plot_data.set_data('e_float', [])
+    self.plot_data.set_data('d_float', [])
     self.plot_data.set_data('t', [])
     self.plot_data.set_data('curr_n', [])
     self.plot_data.set_data('curr_e', [])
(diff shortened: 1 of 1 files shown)


denniszollo:
I'd rather do the dual color thing

henryhallam:
2015-07-06-141832_1520x882_scrot

Crosshair is now only shown for fixed solutions, not float ones.

cbeighley:
I'd vote for the crosshair being shown on both.

piksi_tools/console/baseline_view.py
```diff
@@ -208,23 +215,36 @@ def baseline_callback(self, sbp_msg):
...
+    self.ds = self.ds[-self.plot_history_max:]
+    self.fixeds = self.fixeds[-self.plot_history_max:]
+
+    self.plot_data.set_data('n_fixed', np.array(self.ns)[np.where(self.fixeds)])
+    self.plot_data.set_data('e_fixed', np.array(self.es)[np.where(self.fixeds)])
+    self.plot_data.set_data('d_fixed', np.array(self.ds)[np.where(self.fixeds)])
+    floats = np.logical_not(self.fixeds)
+    self.plot_data.set_data('n_float', np.array(self.ns)[floats])
+    self.plot_data.set_data('e_float', np.array(self.es)[floats])
+    self.plot_data.set_data('d_float', np.array(self.ds)[floats])
```
fnoble: I could see this having some performance impact, did you notice any slowdown when you are at `pol_history_max` points?

fnoble:
Apart from the performance question this looks good, no worse than the existing code at least :)

piksi_tools/console/baseline_view.py
henryhallam: I didn't, but I was testing on a PC with plenty of CPU.  I'll run it with htop on the Atom tablet and see how it looks.

henryhallam:
@cbeighley do you think the crosshair has to change color or are you OK with it being black for both fixed and float?

fnoble:
@cbeighley : do you think it could be confusing to have the corsshair on both? I imagine someone looking at the float crosshair that isn't updating and thinking something was wrong

fnoble:
c.f. RTKLIB, they only use one crosshair only it changes colour

fnoble:
I we do have two though, I think they should be coloured the same colour as their traces (like the SPP tab)

cbeighley:
Don't mind if it's two colors or one, but I think it's good to have an obvious indicator of where the last float solution was.

henryhallam:
Review appreciated
Performance seems OK but could always use more testing
2015-07-09-150153_956x1076_scrot
2015-07-09-150137_956x1076_scrot

cbeighley:
Looks good, you want me to bench test?

henryhallam:
Sure if you have a chance to run it on a crappy computer, that'd be great.  But honestly I really don't expect this change to have any kind of noticeable performance impact.

henryhallam:
Made quick quantitative CPU usage comparison, doesn't appear to be a difference although either way, the console is a CPU hog.

PR #108, opened 2015-07-08
update expert settings for new settings messages

denniszollo:
cc @henryhallam .  The new settings messages broke the expert stuff on new firmware.


henryhallam:
Thanks for fix!

PR #113, opened 2015-07-10
Bootloader integration tests (rebased)

cbeighley:
Finally! Rebased bootloader-tests branch.

This adds an integration test suite for the bootloader, `tests/test_bootloader.py`, and a shell script to orchestrate that over bootloader / firmware releases, `piksi_tools/testing/run_bootloader_tests.sh`, using the aforementioned script and a Black Magic Probe. The latter still needs to get deployed on Jenkins. Still need to add tests for when Piksies are connected to eachother over UART.

Testing script has been used to test latest piksi_bootloader release candidate (it passes).

Other additions: `piksi_tools/heartbeat.py`, `piksi_tools/utils.py`, `piksi_tools/timeout.py`. The latter uses signal and won't work on Windows, but is only meant for integration tests. Also miscellaneous and sundry changes to get working with latest libsbp.

/cc @mfine @mookerji @fnoble


tests/test_bootloader.py
```diff
@@ -0,0 +1,634 @@
...
+
+    Parameters
+    ==========
+    handler : sbp.client.handler.Handler
+      Handler to send messages to Piksi over and register callbacks with.
+    """
+
+    self.settings_received = False
+
+    with SettingsView(handler,
```
mfine: Maybe not in this PR, but we should move this on to `piksi_tools.Diagnostics` and a `Diagnostics(link).diagnostics` call.

tests/test_bootloader.py
```diff
@@ -0,0 +1,634 @@
...
+          print "Piksi Bootloader Version:", piksi_bootloader.version
+
+        # Wait for heartbeat, get settings, print firmware/NAP versions.
+        heartbeat = Heartbeat()
+        handler.add_callback(heartbeat, SBP_MSG_HEARTBEAT)
+        if self.verbose: print "Waiting to receive heartbeat"
+        while not heartbeat.received:
+          time.sleep(0.1)
+        if self.verbose: print "Received hearbeat"
+        handler.remove_callback(heartbeat, SBP_MSG_HEARTBEAT)
```
mfine: These add_callback, remove_callback's could be moved on to `handler.wait` and clean up the code.

mfine:
Nice, tests look good and good useage of messages from `libsbp`. A couple of areas could be improved: using Diagnostics instead of SettingsView, and using link.wait calls (I don't think they were working when this first got started :(). But looks ok to me.

tests/test_bootloader.py
cbeighley: Not sure I follow how that can be done - doesn't look like `handler.wait` can be passed a callback or something?

mfine: Hmm, right - wait will just give you back data... we should have a `wait_callback` that allows you to pass a callback in, too.

PR #128, opened 2015-07-20
Move to deprecated msg print

mfine:
Deprecated print, new sbp.


mfine:
Going to bring this in.

PR #132, opened 2015-07-22
Runner for USRP HITL test.

mookerji:
Command line script for running test USRP HITL test scenarios.

/cc @henryhallam @imh @mfine @denniszollo @cbeighley @fnoble @gsmcmullin

### DIFF
--- a/piksi_tools/testing/usrp_test.sh
+++ b/piksi_tools/testing/usrp_test.sh
@@ -1,0 +1,208 @@
+#!/bin/bash -vxe
+# Copyright (C) 2015 Swift Navigation Inc.
+# Contact: Bhaskar Mookerji <mookerji@swiftnav.com>
+#          Henry Hallam <henry@swiftnav.com>
+#
+# This source is subject to the license found in the file 'LICENSE' which must
+# be be distributed together with this source. All other rights reserved.
+#
+# THIS CODE AND INFORMATION IS PROVIDED "AS IS" WITHOUT WARRANTY OF ANY KIND,
+# EITHER EXPRESSED OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE IMPLIED
+# WARRANTIES OF MERCHANTABILITY AND/OR FITNESS FOR A PARTICULAR PURPOSE
+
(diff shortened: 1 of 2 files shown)


henryhallam:
I think when you start the playback via ssh you can just leave that synchronous and blocking until it completes.

piksi_tools/testing/usrp_test.sh
```diff
@@ -0,0 +1,207 @@
...
+    run_parallel_pairs record_pair
+    start_scenario
+    echo "Done!"
+}
+
+function show_help () {
+    echo "Move along, nothing to see here..."
+}
+
+while getopts ":x:" opt; do
```
gsmcmullin: Why bother with all this arg stuff when you can only get it wrong?  The script can only do one thing, why not just do it?

piksi_tools/testing/Makefile
```diff
@@ -7,3 +6,5 @@ hub_ctrl: hub_ctrl.o
 clean:
 	rm -f *.o hub_ctrl
 
+usrp_test:
```
gsmcmullin: I don't think this adds any value.  It doesn't build anything, and it's more obvious to just run the script than grep the makefile for the target.

requirements.txt
```diff
@@ -9,4 +9,4 @@ pytest-cov==1.8.1
 tox==1.8.1
 virtualenv==1.11.6
 intelhex
-
+pyudev
```
gsmcmullin: This appears to be unrelated.

piksi_tools/testing/usrp_test.sh
mookerji: If I broke into two parts (1. program., 2., run serial capture, etc.) would it make more sense then?

requirements.txt
mookerji: It's for cycling the hub, but it should probably be moved out of requirements and into the script.

piksi_tools/testing/Makefile
mookerji: Removed.

requirements.txt
mookerji: Removed.

PR #149, opened 2015-07-29
Mavlink decode and Ardupilot subdirectory

denniszollo:
This adds an ability to convert the binary file logged on Ardupilot into JSON serialized SBP.  It also moves ardupilot related code into a subdirectory. This also fixes some problems from a similar file I sent to eli and Neils via Gist.
cc @njoubert @mookerji @mfine @fnoble @eliwu26


piksi_tools/ardupilot/mavlink_decode.py
```diff
@@ -0,0 +1,141 @@
...
+      sender_id = last_m.sender_id
+      bin_data = bytearray(last_m.get_raw_msgbuf()[SBR1_DATASTART:SBR1_DATASTART+msg_len])
+      last_m = m
+    else:
+      #just save this one
+      last_m = m
+    if bin_data != None:
+      if(len(bin_data) != msg_len):
+        print ("Length of SBP message inconsitent for msg_type {0}. "
+          "Expected Lenght {1}, Actual Length {2}").format(msg_type, msg_len, len(bin_data))
```
mookerji:
```
print "Length of SBP message inconsistent for msg_type %d. " % msg_type
print "Expected Length %d, Actual Length %d" % (msg_type, msg_len, len(bin_data)
```

piksi_tools/ardupilot/mavlink_decode.py
```diff
@@ -0,0 +1,141 @@
...
+      extractedData.append((timestamp, msg_type, sender_id, msg_len, bin_data))
+      num_msgs+=1
```
mookerji: nit: `num_msgs += 1`

piksi_tools/ardupilot/mavlink_decode.py
```diff
@@ -0,0 +1,141 @@
...
+  print "extracted {0} messages".format(num_msgs)
+  return extractedData
+
+def rewrite(records, outfile):
+  """returns array of (time delta offset from beginning of log in msec,
```
mookerji: nit: Returns

piksi_tools/ardupilot/mavlink_decode.py
```diff
@@ -0,0 +1,141 @@
...
+  extractedData = []
+  log = DFReader_binary(filename)
+  log.Verbose = True
+  last_m = None
+  num_msgs=0
+  while True:
+    m = log.recv_match(type=['SBR1', 'SBR2'])
+    if m is None:
+        break
+    m.Verbose = True
```
mookerji: nit: verbosity should be optional.

piksi_tools/ardupilot/mavlink_decode.py
```diff
@@ -0,0 +1,141 @@
...
+
+This decoder requires commit 0b6e5ab1f6d7911d408aaee8a4ec7a457e238399
+which is currentl in the denniszollo fork on Github as a pull request
+'''
+def extractSBP(filename):
+  extractedData = []
+  log = DFReader_binary(filename)
+  log.Verbose = True
+  last_m = None
+  num_msgs=0
```
mookerji: nit: `num_msgs = 0`.

piksi_tools/ardupilot/mavlink_decode.py
```diff
@@ -0,0 +1,141 @@
...
+The bytearray contains raw SBP binary data logged directly from the serial port.
+Each tuple should contain exactly one SBP message.
+
+This decoder requires commit 0b6e5ab1f6d7911d408aaee8a4ec7a457e238399
+which is currentl in the denniszollo fork on Github as a pull request
+'''
+def extractSBP(filename):
+  extractedData = []
+  log = DFReader_binary(filename)
+  log.Verbose = True
```
mookerji: nit: `log.Verbose = True`  should be optional (set as an argument to `extractSBP`.

piksi_tools/ardupilot/mavlink_decode.py
```diff
@@ -0,0 +1,141 @@
...
+'''
+This function takes in a filename for a ArduPilot dataflash log,
+and returns an array of (timestamp, bytearray) tuples.
+The bytearray contains raw SBP binary data logged directly from the serial port.
+Each tuple should contain exactly one SBP message.
+
+This decoder requires commit 0b6e5ab1f6d7911d408aaee8a4ec7a457e238399
+which is currentl in the denniszollo fork on Github as a pull request
+'''
+def extractSBP(filename):
```
mookerji: nit: there ought to be a test with a mavlink binary data file in the tests directory.

piksi_tools/ardupilot/mavlink_decode.py
```diff
@@ -0,0 +1,141 @@
...
+from sbp.msg import SBP
+
+import time
+import binascii
+import json
+
+SBR1_DATASTART = 16
+SBR2_DATASTART = 13
+
+'''
```
mookerji: nit: double quote instead of single quote. also after the def line.

piksi_tools/ardupilot/mavlink_decode.py
```diff
@@ -0,0 +1,141 @@
...
```
mookerji: nit: double quote instead of single quote

piksi_tools/ardupilot/mavlink_decode.py
```diff
@@ -0,0 +1,141 @@
...
+  print "extracted {0} messages".format(num_msgs)
+  return extractedData
+
+def rewrite(records, outfile):
+  """returns array of (time delta offset from beginning of log in msec,
+  timestamp in sec, SBP object, parsed). skips unparseable objects.
+
+  """
+  new_datafile = open(outfile, 'w')
+  protocol = 2
```
mookerji: `protocol` here isn't used, since the cPickle code is gone.

piksi_tools/ardupilot/mavlink_decode.py
```diff
@@ -0,0 +1,141 @@
...
+SBR1_DATASTART = 16
+SBR2_DATASTART = 13
+
+'''
+This function takes in a filename for a ArduPilot dataflash log,
+and returns an array of (timestamp, bytearray) tuples.
+The bytearray contains raw SBP binary data logged directly from the serial port.
+Each tuple should contain exactly one SBP message.
+
+This decoder requires commit 0b6e5ab1f6d7911d408aaee8a4ec7a457e238399
```
mookerji: Does it still require that commit?

piksi_tools/ardupilot/mavlink_decode.py
```diff
@@ -0,0 +1,141 @@
...
+    if m is None:
+        break
+    m.Verbose = True
+    bin_data = None
+    timestamp = None
+    msg_type = None
+    sender_id = None
+    msg_len = None
+    if last_m != None and last_m.get_type() == 'SBR1' and m.get_type() == 'SBR2':
+      #append the two
```
mookerji: When we were originally working on this, it was unclear how this parser actually worked (because the mavlink format was not understood). If you could possibly add some comments on the actual control flow here, that would be helpful.

piksi_tools/ardupilot/mavlink_decode.py
```diff
@@ -0,0 +1,141 @@
...
+      timestamp = getattr(last_m, '_timestamp', 0.0)
+      msg_type = last_m.msg_type
+      sender_id = last_m.sender_id
+      bin_data = bytearray(last_m.get_raw_msgbuf()[SBR1_DATASTART:SBR1_DATASTART+msg_len])
+      last_m = m
+    else:
+      #just save this one
+      last_m = m
+    if bin_data != None:
+      if(len(bin_data) != msg_len):
```
mookerji: nit: parens not needed in the `if` clause here.

piksi_tools/ardupilot/mavlink_decode.py
denniszollo: I have a PR up on mavlink to allow you to get the binary directly without first parsing it and recreating.  It hasn't been commented on or merged, so for now you have to check out my branch or cherry pick the commit.  It might work without this, but it will be much harder.

piksi_tools/ardupilot/mavlink_decode.py
```diff
@@ -0,0 +1,141 @@
...
+    m = log.recv_match(type=['SBR1', 'SBR2'])
+    if m is None:
+        break
+    m.Verbose = True
+    bin_data = None
+    timestamp = None
+    msg_type = None
+    sender_id = None
+    msg_len = None
+    if last_m != None and last_m.get_type() == 'SBR1' and m.get_type() == 'SBR2':
```
mookerji: at a higher-level, you might consider breaking this parsing bit into an actual iterator pattern (using `yield`, etc.).

PR #151, opened 2015-07-31
Update libsbp interfaces.

gsmcmullin:
Updates to match swift-nav/libsbp#228.

/cc @mfine @mookerji


mfine:
This is looking great! Nice cleanup!

piksi_tools/action_logger.py
```diff
@@ -254,9 +254,9 @@ def main():
           link.add_callback(sl.printer, SBP_MSG_PRINT_DEP)
```
mookerji: Do we want to apply the Forwarder to the other callbacks (printer, etc.)?

piksi_tools/bootload.py
```diff
@@ -81,7 +81,7 @@ def handshake(self, timeout=None):
       t0 = time.time()
     self.handshake_received = False
     expire = time.time() + 15.0
-    self.link.send(SBP_MSG_RESET, "")
+    self.link(MsgReset())
```
mookerji: nit: Actually, I'm pretty sure there's a `self.link.reset()` that does just this.

piksi_tools/bootload.py
```diff
@@ -92,13 +92,13 @@ def handshake(self, timeout=None):
         self.link.send(SBP_MSG_RESET, "")
```
mookerji: ditto on reset

mookerji:
This is phenomenal.

piksi_tools/action_logger.py
gsmcmullin: We can and should apply the Forwarder wherever the callbacks can block or take time to run.  There is no real gain where they are quick, or just post something to run on the UI thread.

piksi_tools/bootload.py
gsmcmullin: Hmm.  There was, but it disappeared in my work on the iterators and isn't in 0.50.1.  This will have to do for now.

PR #155, opened 2015-08-06
Output list

denniszollo:
Reformats the text blob of messages in the piksi console into a list sorted by date and annotated with log level.
End result looks like this:
image

The Python stderr and stdout are assigned a "console" log level and cannot be filtered out.  Default filter level is Warning, which filters out info and debug messages.

cc @mookerji @fnoble @gsmcmullin 
Any thoughts on this


piksi_tools/console/console.py
```diff
@@ -159,6 +168,7 @@ class SwiftConsole(HasTraits):
   system_monitor_view = Instance(SystemMonitorView)
   settings_view = Instance(SettingsView)
   update_view = Instance(UpdateView)
+  log_level_filter = Enum(list(SYSLOG_LEVELS.itervalues()))
```
denniszollo: Here I get the possible log levels from a dictionary in the output_list class and use Traits ENUM to handle the gui and selection.

piksi_tools/console/console.py
```diff
@@ -223,13 +235,13 @@ class SwiftConsole(HasTraits):
```
denniszollo: We need different callbacks for deprecated msg_prints vs msg_logs.  the msg_prints will have log level inferred, the msg_logs will go directly into the table with the level that comes in.

piksi_tools/console/console.py
```diff
@@ -242,12 +254,20 @@ def ext_event_callback(self, sbp_msg, **metadata):
   def _paused_button_fired(self):
     self.console_output.paused = not self.console_output.paused
```
denniszollo: When the gui's log level filter is changed, we need to change the filter trait on the table view as well.  I couldn't figure out how to have the sub-part of the view re-use the upper level trait.

denniszollo:
@mookerji per our convo - ready for review.

mookerji:
Looking in a bit.

piksi_tools/console/console.py
```diff
@@ -221,15 +233,17 @@ class SwiftConsole(HasTraits):
...
-      self.console_output.write(sbp_msg.payload.encode('ascii', 'ignore'))
+      self.console_output.write_level(sbp_msg.payload.encode('ascii'),
+                                      str_to_log_level(msg.split(':')[0]))
     except UnicodeDecodeError:
       print "Critical Error encoding the serial stream as ascii."
 
   def log_message_callback(self, sbp_msg, **metadata):
     try:
-      self.console_output.write(MsgLog(sbp_msg).text.encode('ascii', 'ignore'))
+      self.console_output.write_level(MsgLog(sbp_msg).text.encode('ascii', 'ignore'), MsgLog(sbp_msg).level)
```
mookerji: What the heck is going on here? Why are we re-encoding the message and writing it out again?

piksi_tools/console/output_list.py
```diff
@@ -0,0 +1,295 @@
...
+
+# Set default filter level
+DEFAULT_LOG_LEVEL_FILTER = "WARNING"
+
+DEFAULT_MAX_LEN = 250
+LOG_LEVEL_TOOLTIP = "Log level sent from the device. "\
+                     "If CONSOLE, row was generated by the Console software."
+TIMESTAMP_TOOLTIP = "Time at which the console software received the message,."
+
+def str_to_log_level(level_str):
```
mookerji: ... so there's a better way to do this:

```
SYSLOG_LEVELS_inverted.get(level_str.lower(), DEFAULT_LOG_LEVEL)
```

piksi_tools/console/output_list.py
```diff
@@ -0,0 +1,295 @@
...
+  """
+  log_level = Int
+  timestamp = Str
+  msg = Str
+
+  # log level string maps the int into a the string via the global ALL_LOG_LEVELS dict
+  # If we can't find the int in the dict, we print "UNKNOWN"
+  log_level_str = Property(fget=lambda self:
+                           ALL_LOG_LEVELS.get(self.log_level, "UNKNOWN"),
+                           depends_on='log_level')
```
mookerji: nit: newline

piksi_tools/console/console.py
```diff
@@ -221,15 +233,17 @@ class SwiftConsole(HasTraits):
     title = CONSOLE_TITLE
   )
 
+
   def print_message_callback(self, sbp_msg, **metadata):
     try:
-      self.console_output.write(sbp_msg.payload.encode('ascii', 'ignore'))
+      self.console_output.write_level(sbp_msg.payload.encode('ascii'),
+                                      str_to_log_level(msg.split(':')[0]))
```
mookerji: Does the callback handling in the console swallow exceptions, print stack trace, and then continue, or does it halt the console? If the latter, swallowing the remaining console messages and printing tracebacks might also be a good idea here.

piksi_tools/console/output_list.py
```diff
@@ -0,0 +1,295 @@
...
+# This maps the log level numbers to a human readable string
+# The unused log levels are commented out of the dict until used
+
+# OTHERLOG_LEVELS will be unmaskable
+OTHERLOG_LEVELS  = {
+                    CONSOLE_LOG_LEVEL: "CONSOLE",
+                    }
+# SYSLOG_LEVELS can be filtered
+
+SYSLOG_LEVELS = {#LOG_EMERG : "EMERG",
```
mookerji: Why are these commented out?

piksi_tools/console/output_list.py
```diff
@@ -0,0 +1,295 @@
...
+# A python stdout or stderr should come in as - 1
+
+CONSOLE_LOG_LEVEL = -1
+DEFAULT_LOG_LEVEL = -2
+
+# This maps the log level numbers to a human readable string
+# The unused log levels are commented out of the dict until used
+
+# OTHERLOG_LEVELS will be unmaskable
+OTHERLOG_LEVELS  = {
```
mookerji: `UNMASKABLE_LEVELS`

piksi_tools/console/output_list.py
```diff
@@ -0,0 +1,295 @@
...
+    Parameters
+    ----------
+    log_level : int
+      Log level on which to filter
+    Returns
+    ----------
+    True if message passes filter
+    False otherewise
+    """
+    if self.log_level <= log_level:
```
mookerji:
```
return self.log_level <= log_level
```

piksi_tools/console/output_list.py
```diff
@@ -0,0 +1,295 @@
...
+    self.log_level = level
+    # remove line breaks from the message
+    self.msg = msg.rstrip('\n')
+    # set timestamp
+    self.timestamp = time.strftime("%b %d %Y %H:%M:%S")
+
+  def matches_log_level_filter(self, log_level):
+    """
+    Function to perform filtering of a message based upon the loglevel passed
+    Parameters
```
mookerji: nit: newline before `Parameters`.

piksi_tools/console/output_list.py
```diff
@@ -0,0 +1,295 @@
...
+    ----------
+    log_level : int
+      Log level on which to filter
+    Returns
```
mookerji: nit: newline before `Returns`

piksi_tools/console/output_list.py
```diff
@@ -0,0 +1,295 @@
...
+  # When the value changes from True to False, self._paused_buffer is copied
+  # back to self.unfiltered_list.
+  paused = Bool(False)
+
+
+
+  def write(self, s):
+    """
+    Write to the lists OutputList as STDOUT or STDERR.
+    This method exist to allow STDERR and STDOUT to be redirected into this
```
mookerji: nit: newline before `This`.

piksi_tools/console/output_list.py
```diff
@@ -0,0 +1,295 @@
...
+
+    Parameters
+    ----------
+    s : List
+      Buffer to append
+    s : LogItem
+      Log Item to add
+    """
+    if len(buffer) > self.max_len:
+      assert (len(buffer) - self.max_len) == 1, "Output list buffer is too long"
```
mookerji: This case needs to be handled explicitly (by flushing the buffer or something), instead of asserting.

piksi_tools/console/output_list.py
```diff
@@ -0,0 +1,295 @@
...
+      buffer.pop()
+    buffer.insert(0, s)
```
mookerji: Also, why do you need to prepend to the list here instead of append? Prepending is not cheap (also probably not necessary) and will get called a lot.

mookerji:
The overall thing is pretty useful, although I recommend looking at some of the things listed.

piksi_tools/console/console.py
denniszollo: With the new iterator, the dispatch wasn't working, so I just casted to a MsgLog

piksi_tools/console/console.py
denniszollo: It continues just fine.  You can set stdout and stderr to any filelike thing with a "write" method.

piksi_tools/console/console.py
mookerji: what does "wasn't working" mean specifically?

piksi_tools/console/output_list.py
denniszollo: For UI considerations.  I didn't want to have too many options in the drop-down filter.

piksi_tools/console/output_list.py
denniszollo: So that the newest message was at the top.  The table gui element I used has no ability to sort the list while displaying.   If I use the table element that has built in sorting, each row was too large to be useful.

Also I don't think that the insert is any worse than appending in python.  I looked it up and the insert is nearly the same mutation as appending.

piksi_tools/console/console.py
denniszollo: The item wasn't a MsgLog, it was just a message.  I'll try without the cast and see if it works today.

piksi_tools/console/output_list.py
denniszollo: Actually, pythons docs say pretty clearly that inserting is expensive so I was wrong.

denniszollo: I could use collections.deque instead.  Any thoughts? https://docs.python.org/2/library/collections.html

denniszollo: I could not get around this.  I know it's inefficient, but I'm limited by the gui library.

PR #156, opened 2015-08-07
Independent update, update view ui improvements

denniszollo:
This builds on Vlad's independent update view stuff and changes the update view so it looks like the screen shot below:  All comments appreciated!

Hint if you add '?w=1' to the end of the url, it will remove whitespace from the DIFF in github.

Still to do in a separate PR:
add in modal dialogs for the progress view so users don't try and use the GUI in the background
prevent closing of the update dialogs
add a mouseover hint about the "erase entire STM" and what it does
check in the folder first before downloading firmwares again

### DIFF
--- a/piksi_tools/console/update_view.py
+++ b/piksi_tools/console/update_view.py
@@ -10,9 +10,8 @@
 # WARRANTIES OF MERCHANTABILITY AND/OR FITNESS FOR A PARTICULAR PURPOSE.
 
 from urllib2 import URLError
-from json import load as jsonload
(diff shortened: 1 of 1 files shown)


denniszollo:
image

denniszollo:
@gsmcmullin @ungureanuvladvictor - anyone want to review?

denniszollo:
Here is link to diff with ignoring whitespace.  https://github.com/swift-nav/piksi_tools/pull/156/files?w=1

PR #157, opened 2015-08-07
Hotfix handler interface.

mookerji:
A few more consumers of the Handle required the Framer update from the
recent refactoring.

/cc @gsmcmullin @denniszollo @mfine


gsmcmullin:
Thanks for fixing this.  Looks good.

PR #158, opened 2015-08-07
Use libsbp message classes instead of struct.unpack.

gsmcmullin:
/cc @mookerji @denniszollo


fnoble:
Awesome, this has been a long time coming!

mookerji:
This is great.

mookerji:
This looks perfect. Merge whenever. 

I will note that we have struct packing code for an Almanac message that's probably still in Piksi tools but hasn't had its fields defined in the libsbp spec. That should also probably be changed at some point.

PR #169, opened 2015-08-27
Console list tweaks

denniszollo:
This PR does the following per feedback from sprint exit on 08/21:

* adds a mouseover on log level filter to help tell people how the filtering works
* adds  a command line argument to console set initial log level filter of console
* removes lines on the tabular view of the console output
* puts newlines which exist for formatting of the log/print message as separate rows in the table

### DIFF
--- a/piksi_tools/console/console.py
+++ b/piksi_tools/console/console.py
@@ -212,7 +212,8 @@ class SwiftConsole(HasTraits):
           Item('clear_button', show_label=False, width=8, height=8),
           Item('', label='Console Log', emphasized=True),
           Spring(),
-          UItem('log_level_filter', style='simple', padding=0, height=8),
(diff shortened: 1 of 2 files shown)


denniszollo:
@cbeighley 
Would you mind reviewing?  This address feedback from team after last sprint about the console list view.

cbeighley:
Tested and looks good to merge. One slight note is that I am seeing a Log Level : Console that is not in the drop down list of log levels, which is slightly confusing.

denniszollo:
the "console" log level I set up so that you could not filter it out because I didn't know where on the severity spectrum it should live.  Maybe it would be better if we did the checkboxes as you suggested

cbeighley:
That works for me. Maybe that could be noted in the hover tip text?

denniszollo:
Noted in hover tip text. Merging this.

PR #172, opened 2015-09-16
remove silly nanosecond thing from messages that are only precise to …

denniszollo:
…the msec.

combing two SBP messages (time and a solution) on your PC assumes a lot and makes the recorded files technically wrong compared to the SBP that they record.

### DIFF
--- a/piksi_tools/console/baseline_view.py
+++ b/piksi_tools/console/baseline_view.py
@@ -157,8 +157,6 @@ class BaselineView(HasTraits):
     dist = np.sqrt(soln.n**2 + soln.e**2 + soln.d**2)
 
     tow = soln.tow * 1e-3
-    if self.nsec is not None:
-      tow += self.nsec * 1e-9
 
     if self.week is not None:
       t = datetime.datetime(1980, 1, 6) + \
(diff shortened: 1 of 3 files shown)


fnoble:
Erm, they are precise to the nanosecond! This is (somewhat) useful information that should not just be discarded.

Its explained in the SBP spec that if you get a GPS_TIME and a solution message where the millisecond field matches then the nanosecond field from the GPS_TIME applies to that message.

denniszollo:
Good to know.  So I guess this PR should be changed to a guard in making sure that we don't add the nanoseconds onto the wrong msec for a particular solution if message sequencing gets interrupted or a GPSTIME message gets dropped. 

Here is an excerpt from an SPP log, where you see the effect of this.

2015-08-04 21:02:19.300000,37.7624434651,-122.3890908114,8.0847,6,0
 23 2015-08-04 21:02:19.400033,37.7624422890,-122.3890822674,10.5882,6,0
 24 2015-08-04 21:02:19.499967,37.7624431705,-122.3890839259,10.7509,6,0
 25 2015-08-04 21:02:19.600033,37.7624411211,-122.3890792790,9.4317,6,0

PR #176, opened 2015-09-30
Piksi tools integration for a networking client.

mookerji:
Adds a base station argument to the console and serial_links
args. This consumes the device DNA as a unique ID, and uses this in
the HTTP request header.

/cc @mfine

### DIFF
--- a/piksi_tools/console/console.py
+++ b/piksi_tools/console/console.py
@@ -10,22 +10,22 @@
 # WARRANTIES OF MERCHANTABILITY AND/OR FITNESS FOR A PARTICULAR PURPOSE.
 
 import os
-import sys
+import piksi_tools.serial_link as s
+import sbp.client as sbpc
 import signal
+import sys
 
-from piksi_tools import serial_link
-import sbp.client
+from piksi_tools.version import VERSION as CONSOLE_VERSION
+from piksi_tools.serial_link import swriter, get_uuid
 from sbp.logging import *
-from sbp.piksi import SBP_MSG_RESET
+from sbp.piksi import SBP_MSG_RESET, MsgReset
+from sbp.client.drivers.network_drivers import HTTPDriver
 from sbp.client.drivers.pyserial_driver import PySerialDriver
 from sbp.client.drivers.pyftdi_driver import PyFTDIDriver
 from sbp.ext_events import *
 
-from piksi_tools.version import VERSION as CONSOLE_VERSION
-
 # Shut chaco up for now
 import warnings
-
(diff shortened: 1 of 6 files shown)


piksi_tools/console/console.py
```diff
@@ -63,12 +64,25 @@ def get_args():
                       help="specify the TraitsUI toolkit to use, either 'wx' or 'qt4'.")
   parser.add_argument('-e', '--expert', action='store_true',
                       help="Show expert settings.")
+  parser.add_argument("-a", "--base",
+                      default=[None], nargs=1,
```
mfine: Should we hard-code a default here?

piksi_tools/console/console.py
```diff
@@ -328,6 +349,34 @@ def update_serial():
...
+        link.add_callback(logger)
+        log_filter = DEFAULT_LOG_LEVEL_FILTER
+        if args.initloglevel[0]:
+          log_filter = args.initloglevel[0]
+        SwiftConsole(link, args.update, log_filter, True).configure_traits()
+  try:
+    os._exit(0)
+  except:
+    pass
+elif port is None and base is None:
```
mfine: Maybe put this first so that the conditions don't nest and you can exit early.

piksi_tools/serial_link.py
```diff
@@ -161,6 +176,74 @@ def log_printer(sbp_msg, **metadata):
...
+  a serial link stream.
+
+  Parameters
+  ----------
+  link : file handle
+
+  """
+  def scallback(sbp_msg, **metadata):
+    link(sbp_msg)
+  return scallback
```
mfine: this looks kind of confusing reading here - had a hard time seeing that the return was with swriter, not with scallback.

mfine:
Couple of nits, but looking ok

mookerji:
Yeah, it's a bit messy.... :(

piksi_tools/console/console.py
mookerji: Yep.

mookerji: Not sure you can remove the condition check `port is None and base`. That's for using the console to read only networked observations. If that condition isn't true, it falls through to the rest of the file.

mfine: I was thinking of:

```python
if port is None and base is None:
  sys.stderr.write("ERROR: No data source specified!")
  sys.exit(1)

if port is None and base:
  ...
```

Isolate the error condition and don't bother attaching it to the other check. Just a style and comprehension thing

mookerji: Ah yeah, that's exactly what I was going to do.

PR #183, opened 2015-10-23
Networking updates.

mookerji:
Actually pass through the base station URI. Make that URI have a sane
default and include a boolean flag for networking stuff.

FYI: We have way too many command-line options now.

/cc @mfine @fnoble


mfine:
:+1:

PR #186, opened 2015-10-23
added a command line argument to bootload for max queued ops

denniszollo:
So we can bootload faster when appropriate. @cbeighley


denniszollo:
Trying to close out PRs before the holiday.   Any issues with this one?  It would be nice to be able to bootload wicked fast on my mac.

/cc @mookerji

mookerji:
Didn't we have serious problems with the new bootloader?

mookerji:
But this is harmless, I suppose, since the default is the same.

PR #187, opened 2015-10-27
Creates interpolate_event_positions.py 

rai-swiftnav:
This addition (Hackathon project) allows a user to interpolate position data of event triggers (MsgExtEvent) from a user or another device. When land surveying, the user can press a button connected to the debug port which logs the TOW of those triggers. In post-processing, this code is a tool to extract the data at those triggers. Another application is remote trigger for cameras or other sensor which can be mounted on a UAV or other vehicle for geo-referencing purposes. The script produces a CSV of the interpolated positions when the event was triggered. 

cc @denniszollo @fnoble

### DIFF
--- a/piksi_tools/interpolate_event_positions.py
+++ b/piksi_tools/interpolate_event_positions.py
@@ -88,7 +88,7 @@ def get_position_parameter(message_type,msg_tow,msg_position, tow, numofmsg, msg
     itt+=1
   return value
 
-def get_trigger_positions(message_type,msg_tow,msgout,numofmsg, msg_horizontal, msg_vertical, msg_depth):
+def get_trigger_positions(message_type,msg_tow,msgout,numofmsg, msg_horizontal, msg_vertical, msg_depth, msg_sats):
   # extracts the position at the trigger tow using interpolation subroutine
   itt=0
   while itt<numofmsg:
@@ -104,16 +104,18 @@ def get_trigger_positions(message_type,msg_tow,msgout,numofmsg, msg_horizontal, 
       right_v= get_position_parameter(message_type,msg_tow,msg_vertical, right, numofmsg, msgout)
       left_d= get_position_parameter(message_type,msg_tow,msg_depth, left, numofmsg, msgout)
       right_d= get_position_parameter(message_type,msg_tow,msg_depth, right, numofmsg, msgout)
+      left_sats = get_position_parameter(message_type, msg_tow, msg_sats, left, numofmsg, msgout) 
 
 
       # interpolate trigger position
       msg_horizontal[itt]=lin_interp(left_h, right_h, left, right, msg_tow[itt])
       msg_vertical[itt]=lin_interp(left_v, right_v, left, right, msg_tow[itt])
       msg_depth[itt]=lin_interp(left_d, right_d, left, right, msg_tow[itt])
+      msg_sats[itt] = left_sats
     itt+=1
   return 
 
-def display_data(message_type,msg_tow,msg_horizontal,msg_vertical,msg_depth,msgtype,outfile, numofmsg,msg_flag):
+def display_data(message_type,msg_tow,msg_horizontal,msg_vertical,msg_depth,msgtype,outfile, numofmsg,msg_flag, msg_sats):
   fout= open(outfile,'wt')
   writer = csv.writer(fout)
   if msgtype == 'MsgBaselineNED' :
@@ -127,7 +129,7 @@ def display_data(message_type,msg_tow,msg_horizontal,msg_vertical,msg_depth,msgt
   itt=0
   while itt<numofmsg:
     if message_type[itt]=="MsgExtEvent" and msg_tow[itt] > 0 :
-      writer.writerow((msg_tow[itt],msg_horizontal[itt],msg_vertical[itt],msg_depth[itt],"0",msg_flag[itt]))
+      writer.writerow((msg_tow[itt],msg_horizontal[itt],msg_vertical[itt],msg_depth[itt],msg_sats[itt],msg_flag[itt]))
     itt+=1
   return 
 
@@ -206,9 +208,9 @@ def write_positions(infile, outfile, msgtype, debouncetime):
         print 'done roll'
         fix_trigger_debounce(message_type, msg_tow, numofmsg, debouncetime)
         print ' done bebounce'
-        get_trigger_positions(message_type,msg_tow,msgtype,numofmsg, msg_horizontal, msg_vertical, msg_depth)
+        get_trigger_positions(message_type,msg_tow,msgtype,numofmsg, msg_horizontal, msg_vertical, msg_depth, msg_sats)
         print 'done interpolation'
-        display_data(message_type,msg_tow,msg_horizontal,msg_vertical,msg_depth,msgtype,outfile, numofmsg,msg_flag)
+        display_data(message_type,msg_tow,msg_horizontal,msg_vertical,msg_depth,msgtype,outfile, numofmsg,msg_flag, msg_sats)
         print 'done outputing data '
 
         return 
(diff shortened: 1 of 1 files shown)


piksi_tools/interpolate_event_positions.py
```diff
@@ -0,0 +1,146 @@
...
+with external event triggers.
+"""
+
+from sbp.client.loggers.json_logger import JSONLogIterator
+from sbp.ext_events import SBP_MSG_EXT_EVENT
+import csv
+
+def lin_interp(o, n, otow, ntow, ttow):
+  """
+  Linearly interpolates distance.
```
denniszollo: You Should probably say that the interpolation is precise to the millisecond in this comment

piksi_tools/interpolate_event_positions.py
```diff
@@ -0,0 +1,146 @@
...
+  outfile : string
+    Output file.
+  """
+
+  with JSONLogIterator(infile) as log:
+    log = log.next()
+    lasttow = 0
+    deltalasttrigger = 0
+    positiondesplayed = 1
+    debouncetime = 1000
```
denniszollo: I think debounce time should be a command line argument.

piksi_tools/interpolate_event_positions.py
```diff
@@ -0,0 +1,146 @@
...
+    Log file to get data from.
+  outfile : string
+    Output file.
+  """
+
+  with JSONLogIterator(infile) as log:
+    log = log.next()
+    lasttow = 0
+    deltalasttrigger = 0
+    positiondesplayed = 1
```
denniszollo: I think you should rename the position displayed variable.  I don't understand its function.

piksi_tools/interpolate_event_positions.py
```diff
@@ -0,0 +1,146 @@
...
+  ----------
+  infile : string
+    Log file to get data from.
+  outfile : string
+    Output file.
+  """
+
+  with JSONLogIterator(infile) as log:
+    log = log.next()
+    lasttow = 0
```
denniszollo: Here you need to explain the significance of "lasttow".  It's a very important variable in the algorithm.  Please also rename as discussed, as "lasttow" confused me about is function.

piksi_tools/interpolate_event_positions.py
```diff
@@ -0,0 +1,146 @@
...
+        hostdelta = metadata['delta']
+        hosttimestamp = metadata['timestamp']
+        
+        if msg.__class__.__name__ == "MsgExtEvent" :
+          deltalasttrigger = msg.tow - lasttow
+          if deltalasttrigger > debouncetime:
+            lasttow = msg.tow
+            positiondesplayed = 0
+
+        if msg.__class__.__name__ == msgtype and positiondesplayed == 0 :
```
denniszollo: each time through the loop the msg has only one type, so you should be able to make these all "else if" statements.  This will make it clearer to the next coder how the control flow works.

piksi_tools/interpolate_event_positions.py
```diff
@@ -0,0 +1,146 @@
...
+          if msgtype == "MsgBaselineECEF" or msgtype == "MsgPosECEF" :
+            if previous_msg.flags == msg.flags: #< interpolates only if lock type didn't change.
```
denniszollo: I think you can pull this flags logic outside the case statement as the "flags" field is common to all the position messages.

piksi_tools/interpolate_event_positions.py
```diff
@@ -0,0 +1,150 @@
...
+          deltalasttrigger = msg.tow - triggertow
+          if deltalasttrigger > debouncetime:
+            triggertow = msg.tow
+            dataout = False
+
+        if msg.__class__.__name__ == msgtype and dataout == False :
+          if msgtype == "MsgBaselineECEF" or msgtype == "MsgPosECEF" :
+            if previous_msg.flags == msg.flags: #< interpolates only if lock type didn't change.
+              trigger_x = lin_interp(previous_msg.x, msg.x, previous_msg.tow, msg.tow, triggertow)
+              trigger_y = lin_interp(previous_msg.y, msg.y, previous_msg.tow,msg.tow, triggertow)
```
denniszollo: you missed some spaces here and below.

piksi_tools/interpolate_event_positions.py
```diff
@@ -0,0 +1,150 @@
...
+            if previous_msg.flags == msg.flags: #< interpolates only if lock type didn't change.
+              trigger_lat = lin_interp(previous_msg.lat ,msg.lat, previous_msg.tow, msg.tow, triggertow)
+              trigger_lon = lin_interp(previous_msg.lon ,msg.lon, previous_msg.tow, msg.tow, triggertow)
+              trigger_height = lin_interp(previous_msg.height, msg.height, previous_msg.tow, msg.tow, triggertow)
+              writer.writerow((triggertow, trigger_lat, trigger_lon, trigger_height, previous_msg.h_accuracy,
+                                previous_msg.v_accuracy, previous_msg.n_sats, msg.flags))
+            else: #< otherwise outputs previous data packet received.
+              writer.writerow((triggertow, previous_msg.lat, previous_msg.lon, previous_msg.height, previous_msg.h_accuracy,
+                                previous_msg.v_accuracy, previous_msg.n_sats, previous_msg.flags))
+          dataout = True
```
denniszollo: Please explain in comments why you are setting dataout at this point.

piksi_tools/interpolate_event_positions.py
```diff
@@ -0,0 +1,150 @@
...
+        if msg.__class__.__name__ == msgtype :
```
denniszollo: please add a comment that says you store the last message of the msgtype of interest at all times to be used for interpolation.

piksi_tools/interpolate_event_positions.py
```diff
@@ -0,0 +1,150 @@
...
+    distance prior to trigger
+  otow : integer
+    TOW of data packet previous to trigger
+  ntow : integer
+    TOW of data packet after trigger
+  ttow : integer
+    TOW of trigger 
+  """
+  d = float(n-o)
+  t = (ntow-otow)
```
denniszollo: You need to check the assumptions here about the order of otow, ntow, and tow.  You can consider doing an "assert" statement that otow < ttow < ntow.  This shouldn't happen but is worth checking.  Also might be worth checking against a maximum interpolation time and warning the user.  For instance, maybe we should just error out if the ntow - otow is greater than a few seconds.

piksi_tools/interpolate_event_positions.py
```diff
@@ -0,0 +1,246 @@
...
+    if message_type[itt]=="MsgExtEvent" and msg_tow[itt] > 0 :
+      writer.writerow((msg_tow[itt],msg_horizontal[itt],msg_vertical[itt],msg_depth[itt],"0",msg_flag[itt]))
+    itt+=1
+  return 
+
+
+
+
+
+def write_positions(infile, outfile, msgtype, debouncetime):
```
denniszollo: Change this to "collect_positions" a function that gets all the trigger positions and puts them as a list of interpolated dictionaries for each message. For example:  if there were 2 msg_ext_events, and you wanted the baseline message, you'd get [{tow: 10000, n: 1, e:2, d:3}, {tow: 10010, n: 1, e:2, d:3}].  This way we can use the intermediate output programatically.

Then we'll make a separate library function that writes a list of dictionaries to CSV, which is basically built into the CSV module.

piksi_tools/interpolate_event_positions.py
```diff
@@ -0,0 +1,246 @@
...
+            msg_horizontal.append("0")
+            msg_vertical.append("0")
+            msg_depth.append("0")
+            msg_sats.append("0")
+          numofmsg+=1
+
+      except StopIteration:
+        print "reached end of file after {0} seconds".format(hostdelta)
+        fix_trigger_rollover(message_type, msg_tow, numofmsg)
+        print 'done roll'
```
denniszollo: make these messages more informative to the user.  For example  "processed for rollover error in Piksi event trigger"

PR #188, opened 2015-10-28
updates to build scripts; move more things into requirements.txt

denniszollo:
Anyone want to test this out with a virgin mac for me?
See http://docs.swiftnav.com/wiki/HOW-TO:_Running_the_Piksi_Console_from_source.  we will remove the step about pip install -r requirements.txt if this works
cc @JoshuaGross  @wltr @benjamin0 @sashamitelman


denniszollo:
cc @jacobmcnamee

PR #191, opened 2015-10-29
Fix `tox` test for clean install

JoshuaGross:
Presumably, `tox` should run on a fresh install / a user should be able to follow the README verbatim without having to wrestle with it too much. Thus, bootloader tests are disabled for pytest runs (per @mookerji's suggestion) and PYTHONPATH is included in the readme line.


tests/test_bootloader.py
```diff
@@ -16,6 +16,9 @@
 if os.environ.get('TRAVIS'):
   import pytest
   pytest.skip("Skipping TestBootloader, as we're running in Travis")
+else:
+  import pytest
+  pytest.skip("Skipping TestBootloader, pytest should not run this test.")
```
mookerji: Don't you just want ....

```
import pytest
pytest.skip("Skipping TestBootloader, pytest should not run this test.")
```

JoshuaGross:
@mookerji  fixed

PR #192, opened 2015-10-30
Replace Ansible with simpler bash script.

mookerji:
The Ansible script was a pain, so I've rolled the Debian and OX
installers into a single Bash library that can be invoked from
make. This will hopefully easier to understand and debug by technical
people who don't give a damn about webdev tools. Also:

1. Updates and includes more dependencies in the requirements.txt.
2. Use Travis to test provisioning on OS X and Debian/Ubuntu.

/cc @mfine @fnoble @JoshuaGross


mfine:
Looks good!

PR #195, opened 2015-11-04
Modify get_driver to exit with return code upon failure.

denniszollo:
If we don't have a device, it shouldn't fail silently.
Exit code 1 will be returned if device is missing.  This will help any downstream users of get_driver (bootload, serial_link, etc)

cc @fnoble @mfine


fnoble:
lgtm

mfine:
:+1:

PR #196, opened 2015-11-11
implement timeout on bootload

denniszollo:
We should timeout if we don't receive the handshake from the device after a reasonable amount of time.


fnoble:
lgtm

PR #199, opened 2015-11-13
[do not merge] update console for signal_t changes

jacobmcnamee:

PR #207, opened 2015-11-17
prn vs sid vs sid.sat vs .... ?

mookerji:
Fix satellite identifier issue from libsbp change.

/cc @fnoble @jacobmcnamee


jacobmcnamee:
I have console changes in #199. I think we should check for prn as well as sid.sat to retain support for legacy message types.

mookerji:
Removed deprecated message types.

PR #211, opened 2015-11-20
Console updates for v0.21 firmware

denniszollo:
*decouple heartbeat and threadstate message rate 
*allow a user to flash whatever nap he wants if he presses the flash nap only button
*remove one unnecessary callback in solution view


denniszollo:
@mookerji any issues with these mods to merge in before the console release?

PR #214, opened 2015-11-23
Packaging fixes for console and HOWTO documentation.

mookerji:
As part of getting the OS X console build working with OS X, I made
some fixes that seemed required to get things to build.

0. Absolute package imports.
1. Removing deprecated message callbacks.
2. HOWTO for releases and Makefile update
3. Get travis to check console build

/cc @denniszollo @fnoble @cbeighley @JoshuaGross

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,17 @@ SWIFTNAV_ROOT := $(shell pwd)
 MAKEFLAGS += SWIFTNAV_ROOT=$(SWIFTNAV_ROOT)
 export PYTHONPATH := .
 
-.PHONY: help deps
+ifneq (,$(findstring /cygdrive/,$(PATH)))
+    UNAME := Windows
+else
+ifneq (,$(findstring WINDOWS,$(PATH)))
+    UNAME := Windows
+else
+    UNAME := $(shell uname -s)
+endif
+endif
+
+.PHONY: help deps serial_deps build_console build_console_posix build_console_Darwin build_console_Linux build_console_Windows
 
 help:
 	@echo
@@ -15,8 +25,10 @@ help:
 	@echo "(Please read before using!)"
 	@echo
 	@echo "Please use \`make <target>' where <target> is one of"
-	@echo "  help      to display this help message"
-	@echo "  deps      to install dependencies"
+	@echo "  help           to display this help message"
+	@echo "  deps           to install all dependencies (includes UI deps)"
+	@echo "  serial_deps    to install serial dependencies (no UI deps)"
+	@echo "  build_console  to build the console binary and installer"
 	@echo
 
 all: deps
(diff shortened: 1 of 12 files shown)


piksi_tools/acq_results.py
```diff
@@ -39,7 +39,7 @@ class AcqResults():
   def __init__(self, link):
     self.acqs = []
     self.link = link
-    self.link.add_callback(self._receive_acq_result, [SBP_MSG_ACQ_RESULT, SBP_MSG_ACQ_RESULT_DEP_A])
+    self.link.add_callback(self._receive_acq_result, SBP_MSG_ACQ_RESULT)
```
mfine: this had been important at one time - probably no one cares anymore.

mfine:
Looks good to me if we're ok with yanking support for deprecated messages.

piksi_tools/acq_results.py
```diff
@@ -74,7 +74,7 @@ def mean_max_snrs(self, snr_threshold):
   def _receive_acq_result(self, sbp_msg, **metadata):
     while N_RECORD > 0 and len(self.acqs) >= N_RECORD:
       self.acqs.pop(0)
-    self.acqs.append(MsgAcqResult(sbp_msg) if sbp_msg.msg_type is SBP_MSG_ACQ_RESULT else MsgAcqResultDepA(sbp_msg))
+    self.acqs.append(MsgAcqResult(sbp_msg))
 
   def _receive_acq_result_dep_a(self, sbp_msg, **metadata):
```
denniszollo: why are we removing support for old messages?

piksi_tools/console/baseline_view.py
```diff
@@ -16,7 +16,7 @@
 from enable.api import ComponentEditor
 from enable.savage.trait_defs.ui.svg_button import SVGButton
 from pyface.api import GUI
-from utils import plot_square_axes
+from piksi_tools.console.utils import plot_square_axes
```
denniszollo: What's wrong with the relative imports?  I find it very hard to develop when absolute imports are used and you have a version installed with setup.py on your machine.  Is this where setup.py develop is useful?

mookerji: Addressed Dennis's comment: you need to use `setup.py develop`.

mookerji:
I think we decided that we weren't going to support interfaces for deprecated SBP messages from before the GNSS signal change. This includes the tracking, acquisition, and observation messages. One thing I can do is add in a callback that checks for deprecated message types and warns the user to upgrade. If we do support deprecated message types, we ought to have a cleaner interface (i.e., separate callbacks) instead of dispatching with the callback handler for a selection of messages that have different interfaces.

@fnoble , thoughts?

fnoble:
Agree that we shouldn't support interfaces for deprecated messages. A warning would be a reasonable thing to do.

denniszollo:
Does anyone know why we require cython in requirements.txt.  I fear this is a vestigal  dependency from piksi_firmware

mookerji:
That's a great question, @denniszollo .

I'm going to update this PR with an update prompt and then we should be ready to go.

Makefile
```diff
@@ -26,3 +38,20 @@ deps:
...
+	cd $(SWIFTNAV_ROOT);
+	@echo
+	@echo "Finished! Please check $(SWIFTNAV_ROOT)/piksi_tools/piksi_tools/console/pyinstaller"
+
+build_console_Darwin: build_console_posix
+
+build_console_Linux: build_console_posix
+
+build_console_Windows:
+	@echo "Alas! Makefile setup for console building not supported on Windoze!"
```
denniszollo: I'll plan to setup the windows build through this makefile on windows.

PR #221, opened 2015-12-03
add optional timeout to Diagnostics initializer

scottswift:
adds an optional timeout to `Diagnostics.__init__`
without the timeout, unresponsive piksis will cause the initializer to hang forever
A user who sets the timeout can check
```
    self.settings_received == False
    self.heartbeat_received == False
    self.handshake_received == False
```

to see how far it got

### DIFF
--- a/piksi_tools/diagnostics.py
+++ b/piksi_tools/diagnostics.py
@@ -29,7 +29,7 @@ class Diagnostics(object):
 
   The :class:`Diagnostics` class collects devices diagnostics.
   """
-  def __init__(self, link):
+  def __init__(self, link, timeout=None):
     self.diagnostics = {}
     self.diagnostics['versions'] = {}
     self.diagnostics['settings'] = {}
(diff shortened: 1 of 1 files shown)


piksi_tools/diagnostics.py
```diff
@@ -51,9 +51,12 @@ def __init__(self, link):
                            SBP_MSG_BOOTLOADER_HANDSHAKE_RESP)
     self.link.add_callback(self._print_callback,
                            [SBP_MSG_LOG, SBP_MSG_PRINT_DEP])
+    timeout = time.time() + timeout if timeout is not None else None
     # Wait for the heartbeat
     while not self.heartbeat_received:
       time.sleep(0.1)
+      if timeout is not None and time.time() > timeout:
+        return
```
mfine: might be useful to print something out and/or exit with -1. Looks good otherwise.

scottswift: I don't want the thread to die, I just want to know it didn't work

scottswift: the object already signals what messages it's received

mfine: But that's indirect evidence - which is not as clear to debug if you have to look into the code to see what you timed out on.

scottswift:
forgot about this PR @mfine is it good to merge once I rebase?

mfine:
:shipit:

PR #222, opened 2015-12-06
Console updates for HTTP POST.

mookerji:
0. Updates the HTTP file-like driver to POST messages to Skylark as
well. Doesn't block the UI thread.
1. New Skylark UI view using settings device_id.
2. Removes command line configuration for Skylark usage in console,
since this is now embedded in the console. Moves all this handling to
the networking view, since this is where it obviously belongs.

/cc @mfine @denniszollo @fnoble @JoshuaGross

### DIFF
--- a/piksi_tools/serial_link.py
+++ b/piksi_tools/serial_link.py
@@ -22,6 +22,10 @@ import warnings
 
 from sbp.bootload                       import *
 from sbp.logging                        import *
+from sbp.navigation import SBP_MSG_POS_ECEF, SBP_MSG_POS_LLH, SBP_MSG_BASELINE_ECEF, \
+  SBP_MSG_BASELINE_NED, SBP_MSG_GPS_TIME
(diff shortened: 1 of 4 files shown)


mookerji:
ugh, sorry this took so long. A clean refactoring actually meeting the intended contracts of the Framer/Handler fixed the connection/disconnection freezing from last week. I think it should be ready to go now.

piksi_tools/serial_link.py
```diff
@@ -258,6 +263,16 @@ def run(args, link):
     # SystemExit exception.
     sys.exit(1)
 
+# The default whitelist of messages for the Piksi to publish
+DEFAULT_WHITELIST = [SBP_MSG_HEARTBEAT,
```
mfine: I think we want more messages than this, right? (e.g., uart data for latencies, etc., etc.) Should we either ignore this or move this to a blacklist?

mookerji: Done. Passing through all of them.

PR #229, opened 2015-12-15
add notes regarding the NMEA message rates

denniszollo:
These settings documentation confuses the heck out of users and I get questions all the time.

 I tried to clarify.  Any thoughts on the wording?  Is there a better unit than "tics"? 

cc @soulcmdc @cbeighley


cbeighley:
Looks good.

PR #232, opened 2015-12-21
only display IAR hypothesis count if it is current and not 1. 

### LINKED ISSUE #121: Console - Change the value of "IAR Num. Hyps." when pending IAR
screen shot 2015-07-15 at 12 37 49 pm

Before Piksi has begun IAR, the value of "IAR Num Hyps" (in the Baseline tab of the console) reads "1".  This is confusing, because it suggests that the system has almost finished resolving its hypotheses.

When the system has not yet begun the IAR process, it should not have an integer value... it should have some text such as “Waiting on more common satellites to begin IAR,” or something technically appropriate.

denniszollo:
Otherwise we display the string "None".

This is in the baseline table of the console.  

cc @cbeighley @drewshannon 

Should close #121

### DIFF
--- a/piksi_tools/console/baseline_view.py
+++ b/piksi_tools/console/baseline_view.py
@@ -128,6 +128,7 @@ class BaselineView(HasTraits):
 
   def iar_state_callback(self, sbp_msg, **metadata):
     self.num_hyps = sbp_msg.num_hyps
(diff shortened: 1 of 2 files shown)


ljbade:
Ah nice I found that annoying.

Issue #62, opened 2015-06-11
Put an annotation in the "RTK Position" tab of the console indicating what settings are required for the tab to display information.


denniszollo:

denniszollo:
Or perhaps we hide the tab until the message is received.

sashamitelman:
Hiding the tab may be less than obvious to non-expert users.

Suggest leaving it exposed, but either greying out its contents, displaying a message (ideally in bold or colored text) saying "RTK Fix needed to enable these options," or both.

PR #234, opened 2015-12-22
Rtk position tab annotated

denniszollo:
I added an annotation to the RTK Position tab in the console.  I get questions from users hourly about why this is blank.  Hopefully this message will help point users towards what needs to be done to get the psuedo-absolute position out.  This should close issue #62 

Please let me know if you have thoughts on the wording. (see photo below)

image

/cc @fnoble @mookerji @robhranac @drewshannon


mookerji:
*great*

PR #239, opened 2016-01-16
Uart period

denniszollo:
required to deal with https://github.com/swift-nav/piksi_firmware/pull/600
and
https://github.com/swift-nav/libsbp/pull/329


denniszollo:
Still needs an SBP revision increment in requirements.txt

PR #242, opened 2016-01-29
super simple osx bootstrap script for travis to try

denniszollo:


denniszollo:
no longer relevant as travis mostly works

PR #251, opened 2016-03-11
Add sbp2rinex.py

ljbade:


ljbade:
Still need to fix up the antenna position in the header somehow

ljbade:
Also cycle slips?

PR #262, opened 2016-04-05
Expert toggle

denniszollo:
@ljbade @fnoble 

Add toggle button for expert settings show
image


ljbade:
Just tested it locally.

Only problem is that when you click the toggle you get the annoying "your console is out of date", "your firmware is out of date" pop-ups again when using development versions of console and firmware.

@denniszollo Do you think it is possible to make those pop-up warnings detect when you have a development build (it always has a `v0.**.post**-g*******` pattern).

denniszollo:
-u command line options makes this pop-up go away.  I think I could pretty easily make it ignore development versions too.

ljbade:
Oh wow, wish I had known about `-u` months ago! (and `-e`)

PR #273, opened 2016-05-19
Observation view fixes for new OBS message and doppler computation

denniszollo:
image

requires SBP from https://github.com/swift-nav/libsbp/pull/354


denniszollo:
Closed in favor of  #283 but kept around for resurrection later

PR #288, opened 2016-06-03
Make sure new console works with v0.21 firmware WRT skylark and sbp_r…

denniszollo:
…elay.


mfine:
We should kill these whitelists.

denniszollo:
How do you mean?

denniszollo:
we have very limited bandwidth in the pixhawk use case so only really want to send one type of message.

mfine:
Ok, then remove the white lists from serial link - that has nothing to do with pixhawk, and these white lists are brittle and general not a good idea.

denniszollo:
So you think we have the skylark client send the whole stream instead then?

mfine:
Yup. That's what the other clients are doing.

PR #292, opened 2016-06-18
Sbp2rinex

denniszollo:


ljbade:
Looks good to me.

Have you compared the SBP JSON file and RINEX to verify everything makes it through correctly?

Also we should test a Piksi log against a nearby CORS RINEX log with RTKLIB to verify it can get a fix.

PR #298, opened 2016-06-28
Status bar

asthakeshan:
Still need to implement timeout.


denniszollo:
The PR needs to have the minimal set of changes you need.  I expect you should only have modified solution_view, baseline_view and the console files.

PR #312, opened 2016-08-02
i291 SBP ephe update

pmiettinen:


ljbade:
I think @denniszollo is best to review?

ljbade:
@pmiettinen Its showing a merge conflict

denniszollo:
@pmiettinen.  We need to sequence this correctly as it relies upon a new SBP revision and release which needs to be updated in the requirements.txt of this repository.

denniszollo:
Also, can you help me to understand whether these ephemeris files are expected to be the Rinex "standard or just some intermediate value for your testing?  If these are intended to be Rinex nav files,  we might want to make sure the extension on the .glo ephemeris file corresponds to the typical conventions.

pmiettinen:
@ljbade You probably can answer Dennis' question better ^

ljbade:
@denniszollo certainly the current implementation is not a proper RINEX nav file. I have no idea why they are written they way they are. You would need to talk to person who originally wrote this code as I think @pmiettinen was just following what was already there.

It might be possible to now produce a proper RINEX file.

ljbade:
Hmm looks like @imh originally wrote this stuff.

Personally I think we should rip out this RINEX saving stuff, and instead call out to https://github.com/swift-nav/piksi_tools/pull/292 (which we should merge).

That has much better tested code. Although might need an update to handle L2CM.

Then we can also add proper RINEX nav file support.

denniszollo:
the problem with #292 is it still hasn't been verified.  I tried to get it to handle L2CM.  In either case, I agree we should rip out ephemeris saving right now. @pmiettinen can you revise this just to remove any ephemeris saving?

Thanks in advance!

ljbade:
@denniszollo ah ok, after v3 filter dev has slowed down ill try to finish of the rinex covertor

mfine:
Should be fixed in #337.

PR #337, opened 2016-09-02
New SBP

mfine:
Remove the delta. Pick up the new ephemeris message.

/cc @denniszollo @margaret @pmiettinen


denniszollo:
did you see the import conflict here?
https://travis-ci.org/swift-nav/piksi_tools/jobs/157188023#L1618

I think Josh has run into this before
/cc @jkretzmer

jkretzmer:
yep - this is the result of sbp requiring `construct==2.5.2` put piksi_tools requiring `construct>= 2.5.2`. the requirements for both of these should probably be in sync - is there anything preventing us from using a newer version of construct in libsbp? could also just lock them both to 2.5.2 if there is no compelling reason to upgrade...

mfine:
No compelling reasons! Sounds good to me.

mfine:
Might need to do this to get this working with `crypto`:

https://github.com/pyca/cryptography/blob/master/.travis/install.sh#L70-L87

mfine:
Bringing this in.

PR #343, opened 2016-09-13
FileIO read/write improvements.

gsmcmullin:
- Add support for TCP connections in fileio CLI utility.
- Add write support in fileio CLI utility.
- Rewrite read and write methods to use selective re-transmissions on missing replies before giving up.

### DIFF
--- a/piksi_tools/fileio.py
+++ b/piksi_tools/fileio.py
@@ -11,12 +11,79 @@
 
 import serial_link
 import random
-import array
+import threading
+import time
 
 from sbp.file_io import *
 from sbp.client import *
+from sbp.client.drivers.network_drivers import TCPDriver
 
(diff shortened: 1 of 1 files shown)


piksi_tools/fileio.py
```diff
@@ -195,6 +259,8 @@ def get_args():
   """
   import argparse
   parser = argparse.ArgumentParser(description='Swift Nav File I/O Utility.')
```
denniszollo: I'd love to remove the serial io command line option boilerplate and lift into one file:  Thoughts?
https://github.com/swift-nav/piksi_tools/blob/master/piksi_tools/serial_link.py#L54

Not mandatory and of course there shouldn't be skylark in their probably.

PR #351, opened 2016-10-20
Configurable base and rover device

mfine:
Allow the ability to choose base and rover device settings. Depends on https://github.com/swift-nav/libsbp/pull/388.

/cc @mookerji


ljbade:
sweet

PR #355, opened 2016-10-20
Handle L1P and L2P. Fix bugs in calculating GPS PRN numbers.

ljbade:
Not sure what do with `acq_results.py`. Looks like it needs TLCm but we will be changing that message anyway in https://github.com/swift-nav/libsbp/pull/384


ljbade:
Hmm how do I get the SID `constellation`, seems its undefined?

ljbade:
Tested using Skylark stream:
image

Works.

ljbade:
@denniszollo can you review/merge

denniszollo:
LGTM - merging.

I think we should put these tables of code to signal identifier right into the SBP library when we make the new messages.

ljbade:
@denniszollo yes I agree, we should have enum constants or something

PR #361, opened 2016-11-03
Change from 'SATS' to 'Signals'

pmiettinen:


pmiettinen:
@denniszollo

denniszollo:
Yes we'll put this in.  We need to discuss what makes sense to communicate to users on the wire though in the future.  generic number of signals doesn't give you much information about the constellation.  i.e. you could have 8 signals, but only 4 satellites with a bad DOP.

switanis:
I'd stay with sats.

Issue #327, opened 2016-08-08
Observation view fixes: 


denniszollo:
Correct units spelling on Obs tab (dB-Hz and Hz).
Numbers should always have the same number of decimal places 
align number right.
Logging folder should be that specified by the status bar.

denniszollo:
Closed with sbp 2.0 updates

PR #367, opened 2016-11-23
Hide rinex record, fixed width table font and formatting, add lock

denniszollo:
#327


ljbade:
A few things: add flags column for 1/2 cycle amb.

For the the validity flags the column value should be blank if the value is not valid.

Finally 0 cno means value is invalid so should set blank if zero.

ljbade:
Also should lock time be shown in seconds or Ms?

ljbade:
I can test this for you tomorrow using my new observations firmware PR

denniszollo:
I'm testing on the old obs message; I agree with everything and will update tonight.  I think I will put locktime in raw and converted to seconds.

ljbade:
@denniszollo ah ok

PR #373, opened 2016-11-30
Obs v2.0

denniszollo:
Add in support for new fields, lock time, and flags in latest observation message.

Still backwards compatible with old observation message


ljbade:
@denniszollo awesome, thanks for this

PR #376, opened 2016-12-01
Improve ObservationView handling of SBP 2.x

ljbade:
This tweaks the ObservationView handling of SBP 2.x to make the flags easier to understand.
- adds labels after the flags value to indicate it's meaning, e.g. `PR CP` means pseudorange and carrier phase valid
- sets invalid fields to zero, to make it obvious the value is invalid and should not be interpreted

It also removes the dead RINEX recording code.

@denniszollo Can you review?

### DIFF
--- a/piksi_tools/console/observation_view.py
+++ b/piksi_tools/console/observation_view.py
@@ -17,13 +17,13 @@ from enable.savage.trait_defs.ui.svg_button import SVGButton
 from piksi_tools.console.utils import determine_path,sopen
 from piksi_tools.console.utils import code_to_str, code_is_gps
 from piksi_tools.console.utils import L1CA
+from piksi_tools.console.utils import EMPTY_STR
 
 import os
 import datetime
 import copy
 
 from sbp.observation import *
-
 
 class SimpleAdapter(TabularAdapter):
     columns = [('PRN', 0),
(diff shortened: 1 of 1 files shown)


piksi_tools/console/observation_view.py
```diff
@@ -198,80 +126,88 @@ def obs_packed_callback(self, sbp_msg, **metadata):
...
+      if (flags & 0x08):
+        flags_string += "MD "
+
+      pr = float(o.P) / divisor
+      cp = float(o.L.i) + float(o.L.f) / (1 << 8)
+      cn0 = float(o.cn0) / 4
+
+      # Sets invalid values to zero
+      # TODO(Leith): it would be better if the cell was blank when invalid
+      if sbp_msg.msg_type not in [SBP_MSG_OBS_DEP_A, SBP_MSG_OBS_DEP_B, SBP_MSG_OBS_DEP_C]:
```
denniszollo: Lets send out the raw value even if invalid.  Doesn't help a customer to think that invalid values will always be 0

denniszollo: If we want to hide invalid values, I think we have to change every column to a string and use the "EMPTY_STR" global when no data exists.  That is currently set to be "--".

ljbade: ok

ljbade:
@denniszollo addressed the review comments, can you review again?

piksi_tools/console/observation_view.py
```diff
@@ -36,37 +36,22 @@ class SimpleAdapter(TabularAdapter):
                ('Flags', 7)]
     font='courier'
     alignment='center'
```
denniszollo: we need these format strings in order to prevent round numbers from jumping alignment

denniszollo:
LGTM - merge provided travis passes linux.

PR #378, opened 2016-12-02
tracking.py: add filters for L1 vs L2 tracking

denniszollo:
image

Allows one to reduce clutter on tracking display by showing only l1 or l2 signals
Also tries to make all colors unique.

If a satellite is no longer tracked, its label is removed from the legend


ljbade:
nice

PR #389, opened 2016-12-12
Console based firmware update for Piksi Multi

gsmcmullin:
Changes to the update view to allow updating of Piksi Multi firmware from the console.
The hardware revision read from the Piksi in the settings is used to distinguish between Piksi v2.3.1 and Piksi Multi.  The index.json file is downloaded as before, and support is added for binary firmware files, and the Piksi Multi update mechanism transferring the file to Piksi using SBP FileIO, and committing it to flash with SBP command.

When Piksi Multi is detected, Piksi v2.3.1 only options (erase flash and NAP update) are hidden from the UI:
image

### DIFF
--- a/piksi_tools/console/update_downloader.py
+++ b/piksi_tools/console/update_downloader.py
@@ -43,6 +43,16 @@ class UpdateDownloader:
       raise URLError("Error: Failed to download latest NAP firmware from Swift Navigation's website")
     return filepath
 
+  def download_multi_firmware(self):
+    try:
+      url = self.index['piksi_multi']['fw']['url']
+      filepath = self._download_file_from_url(url)
+    except KeyError:
+      raise KeyError("Error downloading firmware: URL not present in index")
+    except URLError:
+      raise URLError("Error: Failed to download latest NAP firmware from Swift Navigation's website")
+    return filepath
+
   def _download_file_from_url(self, url):
     url = url.encode('ascii')
     urlpath = urlparse(url).path
--- a/piksi_tools/console/update_downloader.py
+++ b/piksi_tools/console/update_downloader.py
@@ -23,9 +23,9 @@ class UpdateDownloader:
     self.index = jsonload(f)
     f.close()
 
-  def download_stm_firmware(self):
+  def download_stm_firmware(self, hwrev):
     try:
-      url = self.index['piksi_v2.3.1']['stm_fw']['url']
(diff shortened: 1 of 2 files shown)


piksi_tools/console/update_downloader.py
```diff
@@ -23,19 +23,29 @@ def __init__(self):
     self.index = jsonload(f)
     f.close()
 
-  def download_stm_firmware(self):
+  def download_stm_firmware(self, hwrev):
     try:
-      url = self.index['piksi_v2.3.1']['stm_fw']['url']
+      url = self.index[hwrev]['stm_fw']['url']
```
denniszollo: This means we will add the STM_FW key into the index file.  I find this a little "wrong" from first principles because there is no "stm" any longer. 

Can we add the switch statement on hwrev here, or should we make the index shema conform?

piksi_tools/console/update_downloader.py
```diff
@@ -23,19 +23,29 @@ def __init__(self):
...
       filepath = self._download_file_from_url(url)
     except KeyError:
       raise KeyError("Error downloading firmware: URL not present in index")
     except URLError:
       raise URLError("Error: Failed to download latest NAP firmware from Swift Navigation's website")
```
denniszollo: can we fix the incorrect message here? (this is in the STM firmware and the message says NAP)

piksi_tools/console/update_view.py
```diff
@@ -509,30 +577,23 @@ def _compare_versions(self):
...
             "New Piksi firmware available.\n\n" + \
             "Please use the Firmware Update tab to update.\n\n" + \
             "Newest STM Version :\n\t%s\n\n" % \
-                self.update_dl.index['piksi_v2.3.1']['stm_fw']['version'] + \
+                self.update_dl.index[self.piksi_hw_rev]['stm_fw']['version'] + \
             "Newest SwiftNAP Version :\n\t%s\n\n" % \
-                self.update_dl.index['piksi_v2.3.1']['nap_fw']['version']
+                self.update_dl.index[self.piksi_hw_rev]['nap_fw']['version']
 
         fw_update_prompt.run()
```
denniszollo: On a slow or nonexistent internet connection it seems like downloading the file makes the console lag when you first start it up.  I think that this is the reason that this version info download was in its own thread.  Can we make it that way again?

jdormoy:
@denniszollo What's left to do on this - can we close it?

PR #419, opened 2017-01-03
Udev rules

denniszollo:
this adds a shell script for configuring udev rules for piksi v2 and piksi multi for user convenience on linux

It also adds the shell script to the pyinstaller package so it will come down with our TARs for linux.

Tested to work on Mufasa (provided the user can sudo).

Would it be better to not have the explicit sudo and make the user figure that out?

/cc @stefan


ljbade:
I think installing udev rules for the user is sensible since the udev config directory is designed to allow packages to install new files into it

At least whenever I have seen new Linux users onboard at Swift they always seem to hit the permission denied issue on the serial device without realising the problem.

PR #447, opened 2017-01-19
USB upgrade UI improvements (master)

cbeighley:
Make instructions more clear and add a console print upon firmware version change.

`parse_version` was just returning None on firmware strings so I removed it. Now firmware version checking and firmware update nagging window works properly.

@denniszollo


denniszollo:
lgtm:  IF linux  and at least one osx build passes on travis, merge at will

cbeighley:
Hallelujah!

PR #458, opened 2017-01-24
Skylark debug

denniszollo:
@jkretzmer @mookerji 

This combined with the write lock on the sbp serial driver is working well for me on the bench.   Please  review if you are able.  Stefan is going to drive this console tomorrow to see if it improves the situation.

Some things to note since you last looked at it.

There is a console command line option to give a yaml string to the skylark view to pre-populate fields in the sbp_relay_view constructor:

That command line options would look like this:
 --skylark  "{'rover_pragma':'trace','rover_uuid':'46f553a3-a0c8-460f-ad3a-f2291d72e6c2', 'verbose':'True'}"

There is also a console command line option to log all stdout/stderr from console to a file: 

  --log-console         Log console stdout/err to file.

The "verbose" option with the --log-console would be intended to help us during drives to get to the bottom of any console skylark issues.

### DIFF
--- a/piksi_tools/console/sbp_relay_view.py
+++ b/piksi_tools/console/sbp_relay_view.py
@@ -108,9 +108,14 @@ class SkylarkWatchdogThread(threading.Thread):
   def stop(self):
     """ stops thread, sets _stop event"""
     self._stop.set()
-    if self.stopped_callback:
-       self.stopped_callback()
     self._stop_time = time.time()
+    if self.stopped_callback:
+      try:
+        self.stopped_callback()
+      except:
+        print "Error stopping SkylarkWatchdogThread: User supplied callback has unhandeled exception") 
+        import traceback
+        print traceback.format_exc()        
     if self.verbose:
       print ("SkylarkWatchdogThread initialized "
              "at {0} and connected since {1} stopped at {2}").format(
--- a/piksi_tools/console/sbp_relay_view.py
+++ b/piksi_tools/console/sbp_relay_view.py
@@ -119,7 +119,7 @@ class SbpRelayView(HasTraits):
 
   def __init__(self, link, device_uid=None, base=DEFAULT_BASE, 
                whitelist=None, rover_pragma='', base_pragma='', rover_uuid='', base_uuid='',
-               connect=False):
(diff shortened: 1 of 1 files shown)


piksi_tools/console/sbp_relay_view.py
```diff
@@ -21,27 +20,206 @@
...
+    return self._connect_time
+
+  def get_stop_time(self):
+    return self._stop_time
+
+  def stop(self):
+    """ stops thread, sets _stop event"""
+    self._stop.set()
+    if self.stopped_callback:
+       self.stopped_callback()
```
JoshuaGross: do we want to wrap this in a try/catch in case the user-code fails?

piksi_tools/console/sbp_relay_view.py
```diff
@@ -21,27 +20,206 @@
...
+    if self.stopped():
+      return 0 # If we stop from the event, it it intended and we return 0
+    else:
+      return -3 # Lost connection
+  
+  def run(self):
+    """ Continuously try and reconnect until thread stopped by other means """
+    while not self.stopped():
+      print "Attempting to connect to skylark..."
+      ret = self.connect(self.link, self.skylark_config)
```
JoshuaGross: should # of reconnections be more stateful, use exponential backoff? For instance, if there's some problem where the server connects initially but disconnects the rover connection immediately after we detect a success, we could rapidly reconnect over and over again.

piksi_tools/console/sbp_relay_view.py
denniszollo: good idea.

piksi_tools/console/sbp_relay_view.py
denniszollo: so you mean sleep 0.1 seconds, then 0.2 seconds, then 0.4 seconds, to avoid hammering it?

denniszollo: I don't think any backoff is needed because the requests library under the hood is doing retries and a 10 second timeout.  If that all fails, we wait 0.25 seconds and retry another round.

PR #459, opened 2017-01-24
Improve tar filename and path

ljbade:
/cc @denniszollo

### DIFF
--- a/piksi_tools/console/pyinstaller/Makefile
+++ b/piksi_tools/console/pyinstaller/Makefile
@@ -17,6 +17,8 @@ else
 endif
 endif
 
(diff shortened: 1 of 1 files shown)


denniszollo:
@ljbade 

Can you make it so it uses UNAME to name the tarfile?  Can you make a corresponding release PR.

ljbade:
@denniszollo looks like uname returns a captial Linux, do you want the file name to have capitals? or should I make it all lowercase?

ljbade:
I decided to add `UNAME_LOWER` which is lower case version on `UNAME` for use in the file name.

PR #471, opened 2017-02-08
Update SBP and move to file handles

mfine:
Pick up latest SBP and move to file handles.

/cc @benjamin0 @denniszollo


benjamin0:
nice, thanks

benjamin0:
@mfine pulled this in, looks like it's working

PR #472, opened 2017-02-08
Settings.yaml : doc updated for master.

switanis:
Several updates to all description.
/cc @denniszollo

### DIFF
--- a/piksi_tools/console/settings.yaml
+++ b/piksi_tools/console/settings.yaml
@@ -6,11 +6,11 @@
   units:
   default value: 'False'
   enumerated possible values: true,false
-  Description: Broadcast surveyed base station location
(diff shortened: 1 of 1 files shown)


piksi_tools/console/settings.yaml
```diff
@@ -228,10 +228,10 @@
 - group: solution
   name: soln_freq
   type: integer
-  units: hz
+  units: Hz
   default value: '10'
   enumerated possible values:
-  Description: The rate at which a solution is generated internally to the Piksi.
+  Description: The frequency rate at which a location solution is computed.
```
denniszollo: "frequency rate" is redundant. Please change to "frequency"

Also, do you mean to say "location" solution as opposed to position solution?

PR #530, opened 2017-03-31
Sbp log zipper

njoubert:
Adds a utility that takes two SBP JSON log files, one from a base station and one from a rover, and interleaves them into a single SBP file. The resulting SBP file can be post-processed using our filter to create a baseline.

### DIFF
--- a/piksi_tools/ardupilot/sbp_log_zipper.py
+++ b/piksi_tools/ardupilot/sbp_log_zipper.py
@@ -1,0 +1,140 @@
+
+import argparse
+import json
+
+import sbp.client.loggers.json_logger as json_logger
+import sbp.observation as ob
+import sbp.navigation as nav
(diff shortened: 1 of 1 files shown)


RaiBearG:
very nice. is the idea to not go the real time route ? log the base, log the rover (on the uav) and get position (rtk) data for triggers or whatever else, in post production ?

njoubert:
@RaiBearG Yes, I'm demoing a post-processing kinematics solution on top of the SBP2 driver. It's a tech demo that will help us move forward on what an end-to-end PPK solution might look like.

denniszollo:
Niels.  Code looks good.  
Can we put some comments and exposition about what it's for?  Suggestion is a doc string on main?
Can we file the template from the other script for the header and copyright stuff?

njoubert:
@denniszollo added!

PR #545, opened 2017-04-20
Disable Skylark UI elements in Networking tab.

mookerji:
Work at https://github.com/swift-nav/piksi_buildroot/pull/197 disables
the need for UI-initiated networking in the upcoming release.

/cc @mfine @denniszollo @jdormoy

### DIFF
--- a/piksi_tools/console/sbp_relay_view.py
+++ b/piksi_tools/console/sbp_relay_view.py
@@ -286,33 +286,6 @@ class SbpRelayView(HasTraits):
                  spring,
                ),
              ),
-             spring,
-             HGroup(
-               VGroup(
-                 HGroup(
-                   spring,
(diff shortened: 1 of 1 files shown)


mookerji:
Looks like this: 

screen shot 2017-04-20 at 4 24 59 pm

Holding off on merging until 1.1 is over.

mfine:
Good with this.

switanis:
Please do not remove Skylark or CORS access from a Console. There are cases when I can only connect RS232 to Piksi.

ljbade:
@stefan we plan to drop support for Skylark via the console for customers on the Skylark Beta. We might be able to hide the button so it only shows in expert mode or something

switanis:
Besides Skylark Beta we also have to do testing and demos. For that Console access is essential.

PR #546, opened 2017-04-21
removed inactive uart info and replaced with network status. 

axlan:
Requires master buildroot


denniszollo:
I actually would like to retain the "latency" portion of the view as this does work and makes sense.

axlan:
Put the measurement latencies back in

PR #549, opened 2017-04-26
Settings_view: Update warning when setting could not be set; prevent …

denniszollo:
…dragging


margaret:
😂 that's good that it was an easy fix

PR #554, opened 2017-04-26
Client Settings

mfine:
Information for client settings for Skylark and NTRIP.

/cc @denniszollo

### DIFF
--- a/piksi_tools/console/settings.yaml
+++ b/piksi_tools/console/settings.yaml
@@ -225,6 +225,26 @@
     bit 2 (decimal value 4) turns on Float IAR simulated RTK outputs
     bit 3 (decimal value 8) turns on Fixed IAR simulated RTK outputs
 
+- group: skylark
+  name: enable
+  type: boolean
+  expert: true
+  units: N/A
+  default value: False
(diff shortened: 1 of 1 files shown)


piksi_tools/console/settings.yaml
```diff
@@ -882,6 +902,26 @@
...
+
+- group: ntrip
+  name: url
+  type: string
+  expert: true
+  units: N/A
+  default value: N/A
+  enumerated possible values:
+  Description: NTRIP URL to use
+  Notes: URL to use with NTRIP client. NTRIP must be enabled to use this setting.
```
denniszollo: Can we explain what the format should be for the url?  at the least I would like to indicate http:// rather than ntrip://

Something like: The URL should take the form "http://XXX.com:PORT_NUMBER/mount_point". For example, http://swift.ntrip.com:2101/sf_office could be a valid ntrip URL

mfine:
Ok, I'll add some text. Full documentation is:

https://curl.haxx.se/docs/manpage.html#URL
http://www.ietf.org/rfc/rfc3986.txt

mfine:
Updated description.

PR #576, opened 2017-05-11
Sbp_relay_view: remove skylark labels, rename variables, remove visib…

denniszollo:
…le default URL

This pull requests is intended as an alternative to #545

Rather than disabling forever, this disables the http connection view by default.  If  a  --networking command line argument is passed to the console, then the view shows up as before.  Optionally, a dictionary can be supplied as an argument to the --networking command line argument which can set any keyword args in the gui element constructor.  In this way we retain the current networking feature from a internet connect PC over serial for testing and demos but avoid confusing our users.   I agree that this code is over-complicated.

Additional notes:
- skylark branding vanquished from UI and variable names
- old "--skylark" command line option is now "--networking"
- old "base" dictionary key is now the more useful "url" key
- when no url is specified, it will default to the new skylark 2 broker
- this will pave the way for a console ntrip client someday in the future along with a refactor of this mess.

Testing:
* bench tested by DZ

/cc @mfine @switanis @mookerji

### DIFF
--- a/piksi_tools/console/sbp_relay_view.py
+++ b/piksi_tools/console/sbp_relay_view.py
@@ -48,15 +48,18 @@ OBS_MSGS = [SBP_MSG_OBS_DEP_C,
           ]
 
 
-class SkylarkConsoleConnectConfig(object):
-  """ This class is intended to encase any skylark specific
+class HttpConsoleConnectConfig(object):
+  """ This class is intended to encase any http specific
       config that will come in.  Eventually should come from a file 
       or simpler command line yaml string. It should hold 
-      skylark api connection info and separate it from the GUI"""
-  def __init__(self, link, device_uid, skylark_url, whitelist,
+      api connection info and separate it from the GUI"""
+  def __init__(self, link, device_uid, url, whitelist,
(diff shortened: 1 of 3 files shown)


mfine:
This is looking ok to me.

piksi_tools/console/sbp_relay_view.py
```diff
@@ -312,12 +316,13 @@ class SbpRelayView(HasTraits):
...
                  spring,
                ),
+             visible_when='show_networking',
              ),
              spring
            )
   )
 
-  def __init__(self, link, device_uid=None, base=DEFAULT_BASE, 
+  def __init__(self, link, show_networking=False, device_uid=None, url='',
```
mookerji: @denniszollo just double check that the caller of this class has updated its arguments, particularly if it depends on the order of the arguments defined here.

mookerji:
See comment, otherwise :shipit:

piksi_tools/console/sbp_relay_view.py
denniszollo: double checked, no dependence on order.

PR #580, opened 2017-05-15
Change git command to not have xcode pop-up in OSX

cbeighley:


piksi_tools/console/settings.yaml
```diff
@@ -695,12 +744,13 @@
...
 - group: ext_events
   name: sensitivity
+  expert: false
   type: integer
   units: us (microseconds)
   default value: 0
   enumerated possible values:
   Description: Minimum time between events (0 = disabled)
-  Notes:
+  Notes: Any event faster than this value will be ignored and a MSG_EXT_EVENT will not be generated.
```
denniszollo: what does faster mean? don't you mean any event that comes n seconds after previous event?

PR #583, opened 2017-05-16
Add GLO support into Trk and Obs views

pmiettinen:
Tested with early version GLO enabled Piksi.

image

image

### DIFF
--- a/piksi_tools/console/utils.py
+++ b/piksi_tools/console/utils.py
@@ -18,10 +18,48 @@ from piksi_tools.utils import sopen
 from sbp.navigation import *
 from threading import Event, Thread
 
-L1CA = 'L1CA'
-L2CM = 'L2CM'
-L1P = 'L1P'
-L2P = 'L2P'
+L1CA = 0
+L2CM = 1
+SBAS_L1CA = 2
+GLO_L1CA = 3
+GLO_L2CA = 4
+L1P = 5
+L2P = 6
+L2CL = 7
+
+SUPPORTED_CODES = [L1CA, L2CM, L2CL, GLO_L1CA, GLO_L2CA]
+
+L1CA_STR = 'L1CA'
+L2CM_STR = 'L2CM'
+SBAS_STR = 'SBAS'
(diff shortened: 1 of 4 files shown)


denniszollo:
This looks good; excited to see GLO in action!

Please move the codefilter class into a utility file of some sort.

Could you also show a screenshot of the tabs operating and mention any testing?

- I was also thinking of eventually extending the filters on tracking view and implementing filters in the obs view to somehow allow filtering by constellation, but we can attack that later.

After we merge into master I can make a windows release build for anyone who needs the binary.

pmiettinen:
@denniszollo I have addressed your comments. Note that I haven't upgraded the color_table for tracking_view, all of the GLO plots are red.

pmiettinen:
Also there's still discussion ongoing about the tracking view legend showing FCN vs slot ID.

piksi_tools/console/utils.py
```diff
@@ -125,10 +171,45 @@ def log_time_strings(week, tow):
...
-    Thread(target=loop).start()    
-    return stopped.set
+  stopped = Event()
+  def loop():
+      while not stopped.wait(interval): # the first call is in `interval` secs
+          func(*args)
+  Thread(target=loop).start()    
+  return stopped.set
+
+
```
denniszollo: Please put this in the gui_utils.py instead of utils file.  The idea is that utils.py has no Traits or UI elements, while gui_utils does.

mbavaro:
@pmiettinen 
Johannes has a PR up on LIBSBP for an updated MSG_TRK_STATE: 
https://github.com/swift-nav/libsbp/pull/469/files
Final decision was:
https://snav.slack.com/archives/C03V0SS40/p1495236589930898

pmiettinen:
Maybe the MSG_TRK_STATE change deserves its own pull request?

denniszollo:
Lets do a separate PR for the new MSG_TRACKING_STATE.   For that one, we should also support both the old message tracking state and the new one for a time.  We can remove any support for older messages if it is in there.

I think this looks good.

PR #589, opened 2017-05-24
Solution cpu rework

pmiettinen:
With this PR

image

Without this PR

image


piksi_tools/console/console.py
```diff
@@ -412,9 +412,9 @@ def update_on_heartbeat(self, sbp_msg, **metadata):
     self.num_sats = temp_num_sats
 
     if self.settings_view: # for auto populating surveyed fields
-      self.settings_view.lat = self.solution_view.latitude
-      self.settings_view.lon = self.solution_view.longitude
-      self.settings_view.alt = self.solution_view.altitude
+      self.settings_view.lat = self.solution_view.latitude_list
```
denniszollo: Are we certain that the  object on the solution_view will never change?  i.e., the nump.pad function doesn't change the underlying object?

piksi_tools/console/solution_view.py
```diff
@@ -127,72 +131,66 @@ def _paused_button_fired(self):
...
-      self.longitude_list = self.longitude_list[-1000:]
-      self.altitude_list = self.altitude_list[-1000:]
-    if len(self.latitude_list) != 0:
-      self.latitude  = sum(self.latitude_list)/len(self.latitude_list)
-      self.altitude  = sum(self.altitude_list)/len(self.latitude_list)
-      self.longitude = sum(self.longitude_list)/len(self.latitude_list)
+    if 0 == self.last_soln.flags:
+      return
+
+    self.latitude_list = np.lib.pad(self.latitude_list,
```
denniszollo: So this doesn't return a different object which would break the link between the settings_view an solution view?

piksi_tools/console/tracking_view.py
```diff
@@ -10,53 +10,53 @@
...
                                       SUPPORTED_CODES
 
+from sbp.tracking import SBP_MSG_TRACKING_STATE, SBP_MSG_TRACKING_STATE_DEP_B
+
 NUM_POINTS = 200
 TRK_RATE = 2.0
+CHANNEL_COUNT = 64
+
+color_code = lambda: int('0x%02x%02x%02x' %
+                         (randint(0,255), randint(0,255), randint(0,255)), 16)
```
denniszollo: This is more elegant, but is it ever useful that PRN23 is the same color every time?

denniszollo:
This looks like a big improvement, but I think the biggest change to CPU load is reducing the drawing frequency to 1hz.

         I'm afraid that drawing only once per second might give the impression that Piksi as a device is slugish. 
         Should we also do this on the baseline tab for completeness?

piksi_tools/console/solution_view.py
pmiettinen: https://docs.scipy.org/doc/numpy/reference/generated/numpy.pad.html says that returned object is always numpy array.

piksi_tools/console/tracking_view.py
pmiettinen: Sure it would bring some value, but at the same time a lot of people seemed to agree that in the original line graph it was very hard to distinguish between separate colors, so the value was lost there.

pmiettinen:
Yes, drawing functions are by far the heaviest here and thus the lowest hanging fruit to pick up.  I agree, I have a feeling that 1 s is a bit too long time to refresh, 0.5 s looks better and doesn't even seem to worsen the performance much.  You decide the final value.

pmiettinen:
I added one more commit and was able to set following refresh rates:
* Tracking 2 Hz | was every time new tracking state message arrived ie. in practice 2 Hz. I think it's better that refresh rate is defined in console as is the case in other tabs
* Solution graph 5 Hz | was 5 Hz
* Solution table follows nav rate | was same
* Baseline graph 5 Hz | was 5 Hz
* Baseline table follows nav rate | was same
* Obs title TOW follows obs rate | was same
* Obs table 2 Hz | was 5 Hz, clear effect on performance but no apparent feeling of sluggishness

CPU figures, same machine as before, but Piksi is running RTK. Earlier pictures are from SPP run. So the circumstances are harder:
image


image


image

pmiettinen:
In previous pictures tab order was out of place, it is fixed now.

piksi_tools/console/tracking_view.py
```diff
@@ -186,7 +262,12 @@ def __init__(self, link):
     self.plot.legend.tools.append(LegendTool(self.plot.legend,
                                   drag_button="right"))
     self.link = link
-    self.link.add_callback(self.tracking_state_callback, SBP_MSG_TRACKING_STATE)
+    self.link.add_callback(self.tracking_state_callback,
```
denniszollo: You can actually specify the messages in a list through the add_callback function:
 self.link.add_callback(self.tracking_state_callback, [SBP_MSG_TRACKING_STATE_DEP_B, SBP_MSG_TRACKING_STATE ] )

denniszollo:
Lastly, can we add the "Tab selected" logic into the IMU view as well so that we don't do any drawing while the tab is not in focus?

jck:
I'm not entirely convinced that drawing occurs on out of focus tabs. On master, when I switch from solution tab to settings tab, the CPU usage goes down considerably.

pmiettinen:
@jck I believe drawing is not done in tabs that are not selected as it's handled by the traitsui(?) but all the calculations are done. For example in obs_packed_callback heavy loops are run all the time even if obs tab is not selected and there's really no need for that since there's no history information to be stored for obs tab.

piksi_tools/console/solution_view.py
pmiettinen: Ok now that I understood what the actual question was I looked into it more detail. So the original solution was that on each call to `update_on_heartbeat()` the lat, lon, alt references in settings_view were updated from solution_view. This of course works but is unnecessary. More matter of taste how you want to do it, I ended up passing solution_view pointer for settings_view. Now settings_view can use the variables from solution_view directly. To be pedant this should be protected by mutex since now there's a possibility that solution_view updates for example longitude_list right after settings_view has read latitude_list. This will cause one sample offset in calculations but since we are calculating mean for static position, effect is negligible.

pmiettinen:
@denniszollo In IMU tab is covered, but note that this tab has the burden of storing limited amount of graph history so the gain is not so great as in obs tab. In IMU tab we can only bypass the `set_data` calls while it's not selected.

denniszollo:
One slight Issue I have is that on OSX the new tabeed methodology looks terrible:
image

pmiettinen:
@denniszollo Can you post a pic how original looks? I'll try to see if I can do anything.. First I need to find OSX first though.

jck:
@pasi the style problem is not related to your changes.

denniszollo:
Turned out the style problem about the tabs was due to dependencies so can be ignored.

silverjam:
@pmiettinen any chance this can be resurrected?  Or this no longer applicable to the current code base?

pmiettinen:
I'll give it a try and see how hopeless it seems.

pmiettinen:
There are some parts that probably can be cherry-picked. For example was it `tracking_view.py` seemed like just easier to start from scratch.

PR #598, opened 2017-06-08
Mavlink decode v2

denniszollo:


piksi_tools/ardupilot/mavlink_decode.py
```diff
@@ -8,92 +8,273 @@
...
+  with open(filename, "rb") as log:
+    while search_binary_key(log, [GPS.HEADER, GPS.HEADER2]):
+      binary = log.read(GPS.SIZE)
+      assert len(binary) == GPS.SIZE, "Length of GPS message inconsistent"
+      gps = GPS(binary)
+      if gps.gwk and gps.gms:
+        return gps
+
+
+def gps_time_to_utc(gps_week, gps_milliseconds, time_us):
```
denniszollo: we are ignoring leapseconds here.  Do you think that matters?  Maybe it would be better to just call it gps_time_to_datetime to avoid an ambiguity.

piksi_tools/ardupilot/mavlink_decode.py
```diff
@@ -8,92 +8,273 @@
...
+TIME_ORIGIN = "1980-01-06T00:00:00.000000"
+
+FORMAT_SIZE_BYTES = {'B': 1,
+                     'H': 2,
+                     'I': 4,
+                     'Q': 8,
+                     }
+
+"""
+Format characters in the format string for mavlink binary log messages
```
denniszollo: can we reference a repo, doc, or source file here for the mavlink format strings?

piksi_tools/ardupilot/mavlink_decode.py
```diff
@@ -8,92 +8,273 @@
...
+  e   : int32_t * 100
+  E   : uint32_t * 100
+  L   : int32_t latitude/longitude
+  M   : uint8_t flight mode
+  q   : int64_t
+  Q   : uint64_t
+"""
+
+
+def search_binary_key(log, searched_keys):
```
denniszollo: Can you add a quick doc string to this function:

Something like:

```
  """ 
  Get a driver based on configuration options

  Parameters
  ----------
  log : string
   description here
  searched_keys : string
    description here
  baud : int
    Serial port baud rate to set.
Returns
  ----------
key: string
  description here
  """
```

piksi_tools/ardupilot/mavlink_decode.py
```diff
@@ -8,92 +8,273 @@
...
+
+  def __init__(self, binary):
+    self.time_us = None
+    self.msg_type = None
+    self.sender_id = None
+    self.msg_len = None
+    self.msg = None
+    self.read_bytes(binary)
+
+  def read_bytes(self, binary):
```
denniszollo: Lets throw an error or information message if the size is incorrect

piksi_tools/ardupilot/mavlink_decode.py
```diff
@@ -118,7 +299,7 @@ def rewrite(records, outfile):
            "data": dispatch(sbp).to_json_dict(),
            "metadata": {}}
```
denniszollo: Current JSON format looks like this:

{"data": {"v_accuracy": 0, "n_sats": 0, "sender": 51763, "msg_type": 522, "lon": 0.0, "tow": 0, "height": 0.0, "crc": 37084, "length": 34, "flags": 0, "h_accuracy": 0, "lat": 0     .0, "preamble": 85, "payload": "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=="}, "session-uid": "9f2c012a-8a1c-44ef-a1ef-c32aef1904ce", "time": "2017-06-08T01:38:47.77178}

Can you make sure this output looks exactly like the output form a json log from the console?

denniszollo: I don't think we need the session_uid, but timestamp should be called "time"

piksi_tools/ardupilot/mavlink_decode.py
denniszollo: I don't think that matters because this just puts some kind of time into the json metadata decorator, but we shouldn't call it UTC because it isn't UTC per say.

piksi_tools/ardupilot/mavlink_decode.py
ebethon: Indeed there is no session-uid and the timestamp is already called "time"

denniszollo:
LGTM

PR #599, opened 2017-06-08
Appveyor + CI enhancements

denniszollo:


jck:
@denniszollo I think this is good to merge now. The only pending task may be to add a note to use `enable==4.6.2` on yosemite

PR #614, opened 2017-06-30
Remove kiwisolver, update deps

jck:
kiwisolver causes build issues for us currently. I noticed that we're not even using it(it is an optional dependency of enable, and we are not using any features which require kiwisolver).

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
-numpy==1.12.1
+numpy==1.13.0
 cython==0.25.2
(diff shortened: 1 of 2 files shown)


ljbade:
curious I wonder if a Kiwi wrote this

ljbade:
although the bird theme is good joke!
"Kiwi is an efficient C++ implementation of the Cassowary constraint solving algorithm"

Cassowary is a large Australian bird, Kiwis are much smaller

PR #616, opened 2017-07-11
Update tracking_view.py

mbavaro:
should we remove that +1?


pmiettinen:
We should retain the +1 in case of the deprecated tracking message.

mbavaro:
sure... but isn't that +1 here?
https://github.com/swift-nav/piksi_tools/blob/master/piksi_tools/console/tracking_view.py#L177

ljbade:
Yes I think it is best to do the +1 in the functions specific to the old messages

mbavaro:
So Leith if you agree I'd merge..

ljbade:
Yes, if you have tested it with both current and the old message

mbavaro:
closing as Keerthan is embracing it!

PR #622, opened 2017-07-27
tracking view and performance improvements

jck:
Done:
* Correct handling, labeling of Glo signals.
* Improve tracking_view performance significantly using `np.roll`

TODO:
* From @pmiettinen 's PR: avoid computations on out of focus tabs
* Glonass colours.


jck:
Moving TODOs to new PR

Issue #636, opened 2017-08-29
Console: drop-down list for COM ports should refresh on click


wurmmi:
in the very first window, starting the console (the interface select):

There's an issue when you open the console first, and plug in the USB serial after that. The new COM port won't show up in the drop-down list. 

Restarting the console/interface-select resolves the issue.

The drop-down list for the COM Ports should refresh when you click on it, in order to increase user-friendliness.

PR #638, opened 2017-08-29
Dynamically refresh serial port list, choose first by default

denniszollo:
Addresses #636

### DIFF
--- a/piksi_tools/console/console.py
+++ b/piksi_tools/console/console.py
@@ -808,7 +808,6 @@ class ShowUsage(HasTraits):
         else:
             self.usage_str = "<pre>" + usage + "</pre>"
 
-
 # If using a device connected to an actual port, then invoke the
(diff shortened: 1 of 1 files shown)


piksi_tools/console/console.py
```diff
@@ -888,15 +887,28 @@ class PortChooser(HasTraits):
...
+        self.refresh_ports()
+        # As default value, use the first city in the list:
+        try:
+            self.port = self.ports[0]
+        except IndexError:
+            pass
+        if baudrate not in BAUD_LIST:
+            self.choose_baud = False
+        self.baudrate = baudrate
+        self.update_ports = call_repeatedly(0.5, self.refresh_ports)
```
mfine: @denniszollo Seems like this isn't letting the process ever exit cleanly on mac os when quiting the selector.

PR #641, opened 2017-08-30
improve console entry point and resource access

jck:


piksi_tools/console/console.py
```diff
@@ -905,89 +853,119 @@ def __init__(self, baudrate=None):
...
+        port = args.port
+        baud = args.baud
+        show_usage = args.help
+        error_str = ""
+    except (ArgumentParserError, argparse.ArgumentError,
+            argparse.ArgumentTypeError) as e:
+        print(e)
+        show_usage = True
+        error_str = "ERROR: " + str(e)
+
```
denniszollo: You can't set the toolkit if you've imported anything from traitsui already

PR #646, opened 2017-09-14
Correct default values in Piksi console

dt-exafore:
# Motivation
See https://github.com/swift-nav/piksi_v3_bug_tracking/issues/674
Following default values mismatch were observed as well (code vs settings.yaml):
- ndb
  > erase almanac: False vs True
  > erase almanac wn: False vs True
  > erase l2c capb: False vs True
  > erase iono: False vs True
  > erase_utc_params: False vs True
- sbp
  > obs msg max size: 255 vs 102
- solution:
  > soln freq: 1 vs 10
  > output every n obs: 2 vs 10
- system info
  > nap channels: 40 vs 24
- uart 0
  > enabled sbp messages: 72,74,65535 vs 68,72,73,74,65535

Default value field can help an user to roll back any settings chages in case user stored new settings in flash, so power cycle does not restore default values.

### DIFF
--- a/piksi_tools/console/settings.yaml
+++ b/piksi_tools/console/settings.yaml
@@ -94,7 +94,7 @@
   expert: false
   type: integer
   units: bytes
-  default value: '102'
+  default value: '255'
   enumerated possible values:
   Description: Determines the maximum message length for raw observation sbp messages
   Notes: This parameter is useful for tuning observation messages for compatibility
@@ -270,7 +270,7 @@
   expert: false
   type: integer
   units: Hz
-  default value: '10'
+  default value: '1'
   enumerated possible values:
   Description: The frequency at which a position solution is computed
   Notes:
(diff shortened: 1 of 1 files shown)


piksi_tools/console/settings.yaml
```diff
@@ -815,7 +815,7 @@
   expert: false
   type: integer
   units: us (microseconds)
-  default value: 0
+  default value: '0'
```
pmiettinen: Is it correct to set this as string if type is integer?

pmiettinen:
Please extract the relevant information from the bug description to this PR description.

piksi_tools/console/settings.yaml
dt-exafore: In other case we will not see the value in Piksi console in Settins tab.
Look at simulator::speed_sigma for example. I tried to do by analogy.

pmiettinen: ok

jck: @denniszollo could you advice about this? (string vs int)

PR #649, opened 2017-09-15
Rename gpgsa_msg_rate to gsa_msg_rate

pmiettinen:
As per https://github.com/swift-nav/piksi_firmware_private/pull/1105


jck:
feel free to merge this after swift-nav/piksi_firmware_private#1105 is merged.

denniszollo:
lgtm.  We should all be aware that the latest console with v1.1 firmware will show a warning; but I think that is preferred to a continually growing list of settings.

PR #652, opened 2017-09-20
serial_link: re-add osx fix to use file rather than CDC driver over usb

denniszollo:
I was able to reproduce the bug where Pyserial opened a garbage serial port for the CDC driver with no data on osx Yosemite.  Since we changed the device ID and Vendor ID for Piksi, we were no longer hitting the code path hack we put in for the v1.0 release.  This minor fix puts us back on the code path to use the file driver rather than whatever pyserial wants to use that isn't reliable.

Testing:
tested extensively on OSX over usb

TODO:
Needs testing on Linux and Windows over USB
we may need to add platform aware logic to get this right

@switanis  when you go to execute the USB on windows test in testlink I'll need to provide you this Appveyor build to test as well as the rc3 console.


denniszollo:
see #407 and #415

PR #658, opened 2017-10-07
Settings confirmation fix

denniszollo:
- This PR does two things, it fixes the bug I introduced when increasing the timeout for the settings reversion functionality 
- It also adds a field to settings.yaml to indicate readonly fields which will not be editable in the gui
- It has been tested on OSX over TCP/IP pretty well

### DIFF
--- a/piksi_tools/console/settings_view.py
+++ b/piksi_tools/console/settings_view.py
@@ -60,11 +60,7 @@ class TimedDelayStoppableThread(threading.Thread):
         return self._stop_event.is_set()
 
     def run(self):
-        start_time = time.time()
-        elapsed_time = time.time() - start_time
-        while elapsed_time < self._delay and not self.stopped():
-            time.sleep(0.2)
-            elapsed_time = time.time() - start_time
+        time.sleep(self._delay)
         if not self.stopped():
             self._target(*self._args)
 
--- a/piksi_tools/console/settings_view.py
+++ b/piksi_tools/console/settings_view.py
@@ -417,7 +417,8 @@ class SettingsView(HasTraits):
                     confirmed_set = False
             if confirmed_set:
                 # If we verify the new values matches our expectation, we cancel the revert thread
-                self.settings[settings_list[0]][settings_list[1]].timed_revert_thread.stop()
+                if self.settings[settings_list[0]][settings_list[1]].timed_revert_thread:
+                    self.settings[settings_list[0]][settings_list[1]].timed_revert_thread.stop()
                 self.settings[settings_list[0]][settings_list[1]].confirmed_set = True
         except KeyError:
             return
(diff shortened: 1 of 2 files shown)


piksi_tools/console/settings_view.py
```diff
@@ -100,44 +131,53 @@ def __init__(self, name, section, value, ordering, settings):
...
+        invalid_setting_prompt.run()
 
     def _value_changed(self, name, old, new):
+        '''When a user changes a value, kick off a timed revert thread to revert it in GUI if no confirmation
+           that the change was successful is received. '''
         if (old != new and old is not Undefined and new is not Undefined):
             if type(self.value) == unicode:
                 self.value = self.value.encode('ascii', 'replace')
+            # Revert_in_progress is a guard to prevent this from running when the revert function changes the value
             if not self.revert_in_progress:
```
jck: does this variable need to be protected with a lock?

piksi_tools/console/settings_view.py
```diff
@@ -42,6 +42,33 @@
...
+        self._stop_event.set()
+
+    def stopped(self):
+        return self._stop_event.is_set()
+
+    def run(self):
+        start_time = time.time()
+        elapsed_time = time.time() - start_time
+        while elapsed_time < self._delay and not self.stopped():
+            time.sleep(0.2)
```
jck: what is the purpose of busy waiting here? can't the thread sleep till timeout elapses or is cancelled by the confirmation?

denniszollo: If I don't check the "stopped" signal, how can I exit the thread early?

piksi_tools/console/settings_view.py
denniszollo: Yes it does, otherwise when the settings value is reverted, it will kick off another thread to confirm the value.  This way I limit the recursion depth to 1.

piksi_tools/console/settings_view.py
denniszollo: Are you suggesting I put a "pass" in to replace the time.sleep()?, I thought that was worse

denniszollo: Oh I see, you would prefer that I just open loop sleep for the delay.  I think that is better and will change.

denniszollo:
untitled
^^ Does this flicker from read only to readwrite upset any apps engineers
@switanis @jkretzmer

jck:
How hard would it be to grey out the value instead of making it disappear?

piksi_tools/console/settings_view.py
denniszollo: I changed this per your suggestion and think it is better now.

denniszollo:
I actually do grey out the value the best I know how, it's just a quirk of the UI that it disappears momentarily.

PR #670, opened 2017-11-01
[WIP] Settings reporting

cbeighley:


piksi_tools/console/settings_report.py
```diff
@@ -0,0 +1,67 @@
...
+        print("post data: uuid is not a string")
+        return
+    # Check data is json.
+    try:
+        json.loads(data)
+    except Exception:
+        print("post data: data is not valid json")
+        return
+
+    data_post = uuid, data
```
denniszollo: I think we need a try catch here to ignore any timeouts or just general connectivity errors.

piksi_tools/console/settings_report.py
```diff
@@ -0,0 +1,67 @@
...
+    r = requests.post('https://catchconsole.swiftnav.com/prof/catchConsole',
+                       headers={'content-type': 'application/json', 'x-amz-docs-region': 'us-east-1'},
+                       data=json.dumps(data_post))
+
+    if r.status_code != requests.codes.ok:
+        print("post data: failed to post data, code:", r.status_code)
```
denniszollo: I don't think we should surface this message to user via print; an offline use case is totally expected.  Perhaps add a "verbosity" flag to the constructor and only print it if we are verbose.  It can follow the verbosity flag for the console itself.

PR #677, opened 2017-12-01
Handle settings write response

gsmcmullin:
For https://github.com/swift-nav/firmware_team_planning/issues/383

For use with https://github.com/swift-nav/piksi_buildroot/pull/397 and https://github.com/swift-nav/piksi_firmware_private/pull/1443


denniszollo:
Before we merge can we also update settings.py cmd line utility as well with same logic?

denniszollo:
replaced by #705

PR #689, opened 2017-12-13
Denniszollo/settings cmd line utility

denniszollo:
Updates to #624 to use an .ini file

### DIFF
--- a/piksi_tools/settings.py
+++ b/piksi_tools/settings.py
@@ -13,6 +13,7 @@ from __future__ import absolute_import, print_function
 
 import struct
 import time
+import yaml
 
 from sbp.client import Framer, Handler
 from sbp.logging import SBP_MSG_LOG, MsgLog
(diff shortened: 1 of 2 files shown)


piksi_tools/settings.py
```diff
@@ -32,76 +66,233 @@ class Settings(object):
...
+    """Write setting by name and confirm set
+
+    Args:
+        section(str): string of section name
+        setting(str): string of setting name
+        value(str): value to set
+        verbose(bool): Echo settings to sdout
+
+    returns:
+        -1 if we were not able to read the setting (may not exist)
```
gsmcmullin: It would be better to just return None and raise an exception on failure

gsmcmullin:
This conflicts heavily with what I was doing for https://github.com/swift-nav/piksi_tools/pull/677
Do you want to merge this first, and then I will rebase, and change it in my PR to use the new response?

denniszollo:
We needed this for a particular customer in short term.  They way I architected it I think it should be very easy to update for the settings_write_resp.

PR #697, opened 2017-12-20
settings.yaml updated to firmware v1.3.9.

switanis:


piksi_tools/console/settings.yaml
```diff
@@ -501,8 +498,8 @@
   default value: N/A
   readonly: True
   enumerated possible values:
-  Description: The UUID of the Piksi.
-  Notes: The UUID is a universally unique identifier for this Piksi.
+  Description: The UUID of the receiver.
+  Notes: The UUID is a Universally Unique IDentifier for this receiver.
```
denniszollo: IDentifier -> Identifier

denniszollo: never mind

PR #698, opened 2017-12-21
Use requests to download fw

denniszollo:


piksi_tools/console/update_downloader.py
```diff
@@ -21,10 +20,14 @@
...
-        self.index = jsonload(f)
-        self.root_dir = ''
-        f.close()
+        try:
+            f = requests.get(INDEX_URL)
+            self.index = f.json()
+            f.raise_for_status()
+            self.root_dir = ''
+            f.close()
+        except requests.ConnectionError as ce:
```
denniszollo: this was required for no internet connection it may be we need to add in other exceptions when this fails.

piksi_tools/console/update_downloader.py
```diff
@@ -62,24 +57,18 @@ def download_multi_firmware(self, hwrev):
...
             raise KeyError(
                 "Error downloading firmware: URL not present in index")
-        except URLError:
-            raise URLError(
-                "Error: Failed to download latest Multi firmware from Swift Navigation's website"
-            )
         return filepath
 
     def _download_file_from_url(self, url):
         if not os.path.isdir(self.root_dir):
```
denniszollo: I changed this exception if the directory doesn't exist; this is a known error case that we try and inform the user about.  IOError is often something we don't predict.

piksi_tools/console/update_view.py
```diff
@@ -376,6 +377,7 @@ def __init__(self,
         self.nap_fw.on_trait_change(self._manage_enables, 'status')
         self.stream = OutputStream()
         self.last_call_fw_version = None
```
denniszollo: I set download_dir to the constructor path

piksi_tools/console/update_view.py
```diff
@@ -506,15 +508,15 @@ def _update_full_firmware_fired(self):
...
+            target=self.manage_firmware_updates, args=("ALL",))
         self._firmware_update_thread.start()
 
     def _download_firmware(self):
         """ Download latest firmware from swiftnav.com. """
         self._write('')
 
         # Check that we received the index file from the website.
-        if self.update_dl is None:
+        if self.update_dl is None or self.update_dl.index is None:
```
denniszollo: If we don't have an index, I skip this whole thing rather than waiting for an attribute error.

piksi_tools/console/update_view.py
```diff
@@ -537,14 +539,25 @@ def _download_firmware(self):
...
-                    "IOError: unable to write to path %s. "
-                    "Verify that the path exists and is writable." %
+                    "RuntimeError: unable to write to path %s. "
+                    "Verify that the path exists." %
                     self.download_directory)
                 self._write("IOError: unable to write to path %s. "
-                            "Verify that the path exists and is writable." %
+                            "Verify that the path exists." %
                             self.download_directory)
+            except IOError as e:
```
denniszollo: I only look for IOErrors we know about to print this error, otherwise I raise

PR #707, opened 2018-01-23
Upgrade view work

denniszollo:
Changes to main upgrade view:
Old:
image
New:
image

Changes to upgrade pop-up text
Old:
image
New:
image

- Fixes bugs where download sometimes failed (especially on OSX) [https://github.com/swift-nav/piksi_v3_bug_tracking/issues/968]
- Fixes bug where the warning to use usb flashdrive upgrade throw error if you launch with command line args
- Removes V2 support from upgrade view which was complicating matters

/cc @switanis @jkretzmer Please provide any comments if you have any

Lastly, you will no longer be able to upgrade a piksi v2 when the PR lands.  You will still get the message that says don't use this version of the console.


denniszollo:
@denniszollo will clean this up, remove mag change dependency, and get it ready for review and merge.

PR #708, opened 2018-01-24
New sig support (try 2)

denniszollo:
- in fulfillment of: https://github.com/swift-nav/piksi_v3_bug_tracking/issues/994
- Has a lot of work arounds to work withthe gui libarary
- Makes us future proof in console for beidu, sbas, and qzss
- Only shows filter for signals that have actually been received in session
- Tested on bench with live signal with GPS and Glonass


denniszollo:
@jck is happy to code review
@jck will test with newer signals after we find him an image or logfile with the newer signals.

PR #710, opened 2018-01-31
Change scaling on Magnetomter view

denniszollo:
Should allow it to show some detail in Earth's field while still not saturating if a magnet is nearby.

/cc @axlan


axlan:
This doesn't saturate from a fridge magnet being within 6 inches?

denniszollo:
I bet it saturates, but as it was rotating the board showed no change on the graph.   Perhaps we just need to dynamically scale instead

denniszollo:
I changed this so that it dynamically scales based upon what has bee received.

axlan:
LGTM

PR #713, opened 2018-02-02
Spectrum analyzer uses FFTMonitor

bryanbrady:
This PR updates the spectrum analyzer view to use libsbp FFTMonitor to reduce code duplication. The code in libsbp FFTMonitor is based on the original spectrum analyzer code.


denniszollo:
This looks good to me; can you rebase on master to make sure that CI passes and verify the version of sbp in the requirements.txt is correct for the FFTMonitor?

PR #718, opened 2018-02-09
Clarify TCP and UDP setting descriptions

ljbade:
First pass at this. Will copy to the other settings entries (e.g. tcp server 1, 2 etc) once wording is approved.


denniszollo:
I think this wording looks great; the one thing I would like to check is that we aren't lying about UDP and the UDP Server actually isn't sending any datagrams.  I was trying to do this with wireshark but am getting wrapped around the axle.

ljbade:
@denniszollo hmm I will be away next week, but ill have a go at wireshark when im back if you haven't managed to test things by then.

ljbade:
@denniszollo can you review this again now that we sorted out the UDP direction.

Then I can merge.

PR #719, opened 2018-02-09
Fix libdrm symbol errors on Ubuntu 16.04

ljbade:
@jck Not sure how to run this to test it

### DIFF
--- a/scripts/build_release.py
+++ b/scripts/build_release.py
@@ -39,6 +39,9 @@ def build_linux():
     maybe_remove(out)
     shutil.move(out_pyi, out)
 
+    # prevents missing export error on ubuntu 16.04 with intel graphics
+    os.remove(os.path.join(out, 'libdrm.so.2'))
+
     print('Creating tar archive')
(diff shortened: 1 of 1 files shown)


scripts/build_release.py
```diff
@@ -39,6 +39,9 @@ def build_linux():
     maybe_remove(out)
     shutil.move(out_pyi, out)
 
+    # prevents missing export error on ubuntu 16.04 with intel graphics
+    maybe_remove(os.path.join(out, 'libdrm.so.2'))
```
jck: looks like you'll need to update the maybe_remove function to let it delete files in addition to directories.

ljbade: bah, do you have some code that will work?

ljbade:
@jck also how do i build the full .tar.gz, or does travis build it for me?

scripts/build_release.py
ljbade: nvm, I just changed it to `os.remove` since it should always exist anyway

ljbade:
@jck where can i find the .tar.gz Travis built?

PR #729, opened 2018-03-01
Updates to settings.yaml for ntrip username/password/gga_out_interval

benjaminaltieri:
Updates to settings per the following buildroot PRs: 
* swift-nav/piksi_buildroot#525
 * swift-nav/piksi_buildroot#527

### DIFF
--- a/piksi_tools/console/settings.yaml
+++ b/piksi_tools/console/settings.yaml
@@ -1392,6 +1392,28 @@
   Notes: If True, NTRIP client will be used.
 
 - group: ntrip
+  name: username
+  expert: false
+  type: string
+  units: N/A
+  default value: N/A
+  readonly: false
+  enumerated possible values:
+  Description: NTRIP username to use. The interface will receive 1002, 1004, 1005, 1006, 1010, 1012, 1033 and 1230 RTCMv3.1 messages and will not transmit or receive any other messages.
+  Notes: Username to use with NTRIP client. NTRIP must be enabled to use this setting.
+
+- group: ntrip
+  name: password
+  expert: false
+  type: string
+  units: N/A
+  default value: N/A
+  readonly: false
+  enumerated possible values:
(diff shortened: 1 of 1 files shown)


piksi_tools/console/settings.yaml
```diff
@@ -1391,6 +1391,28 @@
...
+- group: ntrip
+  name: password
+  expert: false
+  type: string
+  units: N/A
+  default value: N/A
+  readonly: false
+  enumerated possible values:
+  Description: NTRIP password to use. The interface will receive 1002, 1004, 1005, 1006, 1010, 1012, 1033 and 1230 RTCMv3.1 messages and will not transmit or receive any other messages.
+  Notes: Password to use with NTRIP client. NTRIP must be enabled to use this setting.
```
cbeighley: Can you give some guidance as to what special characters are / are not allowed here?

piksi_tools/console/settings.yaml
```diff
@@ -1391,6 +1391,28 @@
...
+- group: ntrip
+  name: username
+  expert: false
+  type: string
+  units: N/A
+  default value: N/A
+  readonly: false
+  enumerated possible values:
+  Description: NTRIP username to use. The interface will receive 1002, 1004, 1005, 1006, 1010, 1012, 1033 and 1230 RTCMv3.1 messages and will not transmit or receive any other messages.
+  Notes: Username to use with NTRIP client. NTRIP must be enabled to use this setting.
```
cbeighley: Can you give some guidance as to what special characters are / are not allowed here?

piksi_tools/console/settings.yaml
benjaminaltieri: I couldn't find any characters the broke it. So no guidance may be better than imprecise guidance.

denniszollo:
Doesn't this PR require a master equivalent? @benjaminaltieri ?

denniszollo:
superseded by #732

PR #733, opened 2018-03-07
Expose Skylark and NTRIP GGA out interval

mfine:
Expose Skylark and NTRIP GGA out interval.

Skylark description probably needs more copy.

/cc @granthausler @denniszollo @cbeighley


silverjam:
change already released

PR #735, opened 2018-03-09
Implement Cell Modem Wizard Usage Widget

benjaminaltieri:
Adds a cellmodem tab and simple usage statistic hardcoded to filter usage on the correct interface (ppp0 for duro is the correct interface)


piksi_tools/console/utils.py
```diff
@@ -237,6 +237,15 @@ def datetime_2_str(datetm):
     return (datetm.strftime('%Y-%m-%d %H:%M'), datetm.strftime('%S.%f'))
 
 
+# Modified based on https://stackoverflow.com/a/1094933
+def sizeof_fmt(num, suffix='B'):
+    for unit in ['', 'K', 'M', 'G', 'T', 'P', 'E', 'Z']:
+        if abs(num) < 1024.0:
+            return "%3.1f%s%s" % (num, unit, suffix)
```
jck: nit: i prefer format strings https://pyformat.info/#simple

benjaminaltieri: good catch and I agree. I will update it

denniszollo:
Can you add a few screenshots?

denniszollo:
Also, why did we make this a separate view; does it really warrant a whole tab?  I think it should be the lower half of the "network view".  Likely the network discovery widget should be on that view as well.

benjaminaltieri:
Oops I thought I was clear to merge.. @denniszollo do you want me to revert? this panel is a feature request per Sabine's document [here](https://docs.google.com/document/d/1KKckvgYPj8wWsEW-HsDbHoc3T7VUjSWphrrepGcFUCY/edit#heading=h.9r3woys1v6b1)

denniszollo:
Lastly, wouldn't it make sense to include the "ifconfig" view for the PPP device?

benjaminaltieri:
Here is a capture, FYI this is a work in progress and additional information will be added to the tab cell_modem_tab_wip

PR #739, opened 2018-03-15
[WIP] Allow display of generic meters north and meters east on solution view

denniszollo:


denniszollo:
same behavior when degrees is chosen:
image
meters put 0,0 at the mean of all lats and longs when meters is first chosen and a scale factor to meters for that location assuming tangent plane there.
image

PR #740, opened 2018-03-16
Add Skylark tab and show serial number and UUID in main window.

gsmcmullin:
For https://github.com/swift-nav/piksi_v3_bug_tracking/issues/1017

screenshot from 2018-03-16 09-59-59

The link is a button as that seems to be the only way to do it in Traits.
The UUID is shown in an editable text box so it can be selected, but if you try to edit it, it just changes back.

### DIFF
--- a/piksi_tools/console/console.py
+++ b/piksi_tools/console/console.py
@@ -191,6 +191,7 @@ class SwiftConsole(HasTraits):
     num_sats = Int(0)
     cnx_desc = Str('')
     latency = Str('')
+    uuid = Str('')
     directory_name = Directory
     json_logging = Bool(True)
     csv_logging = Bool(False)
@@ -356,6 +357,16 @@ class SwiftConsole(HasTraits):
                     ),
                     Item(
                         'latency',
+                        padding=2,
+                        show_label=False,
+                        style='readonly'),
+                    Item(
+                        '',
+                        label='Device UUID:',
+                        emphasized=True,
+                        tooltip='Corrections latency (-1 means no corrections)'
+                    ),                    Item(
+                        'uuid',
                         padding=2,
                         show_label=False,
                         style='readonly'),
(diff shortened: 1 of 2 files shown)


mfine:
Sweet :+1:

granthausler:
Thanks Gareth. Can we change the second sentence to:

"It removes the need for a base station or CORS station"  

(the reasoning here is that we don't want to reveal VRS as part of our service offering and the term is trademarked by trimble)

This all looks great. One other question is how difficult would it be to make another button saying 'Copy UUID' which copies the ID to clipboard when pressed. This is not a hard ask but just a thought.

piksi_tools/console/console.py
```diff
@@ -357,6 +361,16 @@ class SwiftConsole(HasTraits):
...
                         style='readonly'),
+                    Item(
+                        '',
+                        label='Device UUID:',
+                        emphasized=True,
+                        tooltip='Corrections latency (-1 means no corrections)'
+                    ), Item(
+                        'uuid',
+                        padding=2,
+                        show_label=False,
```
denniszollo: If you add editable=False here; then you won't be able to edit

denniszollo:
We might be able to copy to the clipboard through pyside: https://srinikom.github.io/pyside-docs/PySide/QtGui/QClipboard.html

piksi_tools/console/skylark_view.py
```diff
@@ -0,0 +1,61 @@
...
+                    HGroup(spring, Item('link', show_label=False), spring),
+                    Item('uuid', label='Device UUID', width=400),
+                ), spring), spring))
+
+    def __init__(self):
+        self.real_uuid = ''
+
+    def _link_fired(self):
+        webbrowser.open(SKYLARK_URL)
+
```
denniszollo: you won't need any of this below if you set editable=False

denniszollo:
I'm also not thrilled to make another tab because it increased the minimum width of the widget and no longer lets you have a 600 Pixel console

PR #753, opened 2018-04-04
don't attempt to revert if readonly setting changes

jck:
fixes DEVC-379


denniszollo:
LGTM.  This means our read-only metadata in console should be triple checked each release (especially if we start preliferating read-only settings).

PR #756, opened 2018-04-05
add expand_json tool and tests

denniszollo:


silverjam:
@pmiettinen @denniszollo Do we still need this?  I thought Pasi introduced something for this not too long ago.

pmiettinen:
I think this should cover this functionality https://github.com/swift-nav/piksi_tools/blob/master/piksi_tools/sbpjson_expand.py

silverjam:
ok, closing this then

PR #765, opened 2018-04-21
Add support for msg_measurement_t

mbavaro:
What it says, needs libsbp 2.3.16

With the new message, either FCN or SLOT for Glonass satellites can be distributed by FW.
The convention is SLOT if `sid` is [1-28] and FCN if `sid` is in the [100+FCN] range.

This PR changes the console tracking view so that both SLOT and FCN are stored in a dictionary which is indexed by the channel number rather than the FCN as it was before.

This should maintain the user experience and allow both messages to be supported by the console.


mbavaro:
OK let's try to merge...

PR #766, opened 2018-04-21
move cell modem and network widget to network tab

denniszollo:


denniszollo:
Looks like this:
image

I also filtered out the loopback and sit0 interface from the network table, and blanked out the PPP0 tx and rx values which were stuck at 0 and misleading.

PR #822, opened 2018-07-10
Add locks on data between gui thread and SBP thread

denniszollo:


denniszollo:
I think that the crash noticed on v1.6.6 release build of console by Stefan maybe because we aren't doing a good job with thread safety between the SBP thread and the GUI thread, which was made worse when I started doing bulk update of ArrayPlotData rather than one variable at a time.

PR #839, opened 2018-07-25
Add settings documentation entries for the VDF settings.

kmdade:
This is not really important at all, but I figured I would leave this PR up since these features are now present on master in PFWP.


silverjam:
Should these be marked as expert ("Advanced") settings?  That is, add the follow key to the settings blocks:    `expert: true`.

kmdade:
Yes they should, thank you!

PR #841, opened 2018-07-27
[DEVC-963] obs view resize issue; change look of obs counts and filters

denniszollo:
Obs View will look like this now:
image

/cc @srossel


srossel:
Looks pretty good! 
Thank you @dzollo

PR #848, opened 2018-07-31
Copy data out of pending to avoid races [DEVC-975] [v1.6]

silverjam:
If a pending_write is invalidated/completed while we're checking for
time-outs in `_check_pending` we could get into a situation where the
time field is `None`.

To resolve this, copy data so that it can't change out from underneath
us and make sure that potentially racy method calls will succeeded even
if the recod is invalidated.

There may still be races here, but to alleviate this, make sure that
data in record is still in a state where operations on it's fields will
succeed.

### DIFF
--- a/piksi_tools/fileio.py
+++ b/piksi_tools/fileio.py
@@ -11,6 +11,7 @@
 
 from __future__ import absolute_import, print_function
 
+import copy
 import random
 import time
 import threading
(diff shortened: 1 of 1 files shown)


piksi_tools/fileio.py
```diff
@@ -142,13 +165,13 @@ def _window_available(self):
...
         return self
 
     def __exit__(self, type, value, traceback):
-        self.link.remove_callback(self._cb, self.msg_type)
+        self.link.remove_callback(self._write_cb, self.msg_type)
 
-    def _cb(self, msg, **metadata):
+    def _write_cb(self, msg, **metadata):
         for pending_write in self.pending[:]:
             if msg.sequence == pending_write.message.sequence:
```
denniszollo: do we need a try catch around this comparison?

Can we guarantee that pending_write.message.sequence is a float and is not None at this point?

denniszollo:
lgtm if you can answer my one question about whether msg.sequence is guaranteed to be there in the pending queue, (I think yes given you can't initialize the sbp message object without this slot).

piksi_tools/fileio.py
silverjam: I removed the clear of `message` field from `invalidate` so that this comparison would still work and the PendingWrite record only enters `self.pending` after passing through the `track` method which guarantees that the `message` field is set.

PR #856, opened 2018-08-07
fileio: add progress_cb reduction factor to reduce load on GUI during…

denniszollo:
… upgrade

### DIFF
--- a/piksi_tools/fileio.py
+++ b/piksi_tools/fileio.py
@@ -29,7 +29,7 @@ MAX_PAYLOAD_SIZE = 255
 SBP_FILEIO_WINDOW_SIZE = 40
 SBP_FILEIO_TIMEOUT = 5.0
 MINIMUM_RETRIES = 3
-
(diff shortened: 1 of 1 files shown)


piksi_tools/fileio.py
```diff
@@ -346,7 +346,7 @@ def write(self, filename, data, offset=0, trunc=True, progress_cb=None):
                     data='')
                 sr.send(msg)
                 offset += len(chunk)
-                if progress_cb is not None:
+                if progress_cb is not None and seq % PROGRESS_CB_REDUCTION_FACTOR == 0:
```
benjaminaltieri: We could possibly reduce even further if we dynamically calculate `PROGRESS_CB_REDUCTION_FACTOR ` to be 1% of the file size and then even avoid the last cycle where the progress bar would likely 'finish' without actually reaching 100%

denniszollo: I'll add that to my master work on upgrade.

PR #857, opened 2018-08-15
Deque perf fixes crash fixes

denniszollo:
This PR does a number of things:
- It fixes current crashes on master by updating plot data in the "GUI.invoke_later" as per before
- It implement deques on tracking view
- it implements deques on solution and baseline view
- it prevents unnecessary numpy copying / rolling with each SBP callback via deques
- it tries to minimize the number of times we update the ArrayplotData in Chaco across all plots

### DIFF
--- a/piksi_tools/console/baseline_view.py
+++ b/piksi_tools/console/baseline_view.py
@@ -40,7 +40,7 @@ from piksi_tools.utils import sopen
 from .utils import resource_filename
 from .gui_utils import GUI_UPDATE_PERIOD
 
-NUMPOINTS = 1000
+PLOT_HISTORY_MAX = 1000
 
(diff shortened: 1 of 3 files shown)


piksi_tools/console/baseline_view.py
```diff
@@ -391,45 +365,41 @@ def _solution_draw(self):
...
         super(BaselineView, self).__init__()
+        self.pending_draw_modes = []
         self.log_file = None
         self.directory_name_b = dirname
         self.num_hyps = 0
         self.last_hyp_update = 0
         self.last_btime_update = 0
         self.last_soln = None
         self.last_mode = 0
+        self.slns = {'n_fixed': deque(maxlen=NUMPOINTS),
```
jck: can this be plot_history_max?

PR #862, opened 2018-08-29
[DEVC-548] Remove explicit Klobuchar enable/disable setting

pmiettinen:
See https://github.com/swift-nav/piksi_firmware_private/pull/2294

### DIFF
--- a/piksi_tools/console/settings.yaml
+++ b/piksi_tools/console/settings.yaml
@@ -290,9 +290,7 @@
   enumerated possible values: True,False
   Description: Enable Skylark client.
   Notes: |
-      If True, Skylark client will be used. Klobuchar correction must be disabled
-      if order for Skylark to function properly, set solution.disable_klobuchar_correction
-      to True before enabling Skylark.
+      If True, Skylark client will be used for reference observations
 
 - group: skylark
   name: url


denniszollo:
We have tended not to remove settings documentation entries so the console can be backwards compatible with older firmwares.  The console warns you if a setting is not documented.
  
In what firmware will the setting be going away? 

If it is going away in v2.1, we can merge this (and possible remove any other stale settings) right after branching for v2.1 so it would end up in the v2.2 console release.

pmiettinen:
@denniszollo Please see discussion in https://swift-nav.atlassian.net/browse/DEVC-548 , might be that this is not needed at all. But I understand your point and will keep it in mind before and if we get to merging state.

pmiettinen:
@denniszollo any better suggestions how to phrase the notes section so that it would suit both old and the new version?

piksi_tools/console/settings.yaml
```diff
@@ -290,9 +290,8 @@
   enumerated possible values: True,False
   Description: Enable Skylark client.
   Notes: |
-      If True, Skylark client will be used. Klobuchar correction must be disabled
-      if order for Skylark to function properly, set solution.disable_klobuchar_correction
-      to True before enabling Skylark.
+      If True, Skylark client will be used. Klobuchar corrections are disabled in order for Skylark to function properly.
```
benjaminaltieri: If we've removed klobuchar corrections as a setting, can we just remove the references to it? @anth-cole

anth-cole: I'd agree - let's just make this setting say `If True, Skylark client will be used for reference observations`

PR #867, opened 2018-09-05
Change timeout on soln mode status bar [DEVC-1061]

denniszollo:
@pmiettinen.  I think this address DEVC-1061.  Let me know what you think!


pmiettinen:
I tested with this, I can still see `No Fix` for about one second on the status bar. So better, but not quite perfect.

Any way to make the status bar and the solution tab to receive fix type from same function/widget?

denniszollo:
Ok, this is rather complicated, but should behave like this:


- If we have a recent solution tab solution we display mode from the solution tab.  We append "INS" to this string if there is INS in use.
- If we have a recent baseline tab mode that is differential (RTK, Float, or DGPS) while the solution tab mode is not differential, we display baseline mode and baseline n_sats.  This will happen in heading or moving base mode
- If there is no recent update from solution or baseline messages, we display mode "None" and n_sats 0

pmiettinen:
DEVC-1061 is confirmed fixed with this. But those indeed are some nasty looking if clauses o_O

PR #868, opened 2018-09-06
Settings View Updates

denniszollo:
Addressing DEVC-1057 led down a rabbit hole on settings view and the results are represented by this PR. It improves performance of settings_view in the following ways:

- Reads all the settings from device faster (about 5x speedup typically) by reading in batches of 10.  We could really go nuts and read 50 at a time for a zippy experience.
- Fixes DEVC-1057 reported by @pmiettinen by avoiding re-constructing settings when MSG_STARTUP is received but rather by just confirming values when MSG_STARTUP is received.  This is because the contract that MSG_STARTUP happens early on in boot which the old console assumed is no longer met.  This was causing the revert thread from being lost in the ether when self.settings on the view was completely replaced.  There is a slight regression in that a user swapping between firmwares may still see settings from the old firmware that no longer exist upon first booting after upgrade.
- Retries reading settings if lossy messaging layer prevented one of the read_by_index messages from making it to the piksi which was hurting user experience for people in that the settings view was often blank when serial port was overloaded or when the comm layer was loosing info.  For each 10 settings read, it will retry a max of one time.  Otherwise user will have to press the "refresh" button.
- fixes issue settings.yaml for default value on dynamics filter settings not being a string as expected.  All settings are strings so they are encoded in yaml as strings as well.

### DIFF
--- a/piksi_tools/console/console.py
+++ b/piksi_tools/console/console.py
@@ -525,7 +525,7 @@ class SwiftConsole(HasTraits):
         self.forwarder = sbpc.Forwarder(self.link, self.logger)
         self.forwarder.start()
         if self.settings_view:
-            self.settings_view._settings_read_button_fired()
+            self.settings_view._settings_read_by_index()
 
     def _stop_json_logging(self):
         fwd = self.forwarder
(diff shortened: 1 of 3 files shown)


denniszollo:
This has been tested extensively on the bench by @dzollo with MASTER buildroot on a piksi over serial and ethernet.   Everything appears in working order.

PR #869, opened 2018-09-07
Advance hardware info settings in system info group and add product id [DEVC-472]

benjaminaltieri:
Per:
https://swift-nav.atlassian.net/browse/DEVC-472
https://github.com/swift-nav/piksi_buildroot/pull/552#issuecomment-379345790

We want `product_id` to be the customer facing setting for product identification

### DIFF
--- a/piksi_tools/console/settings.yaml
+++ b/piksi_tools/console/settings.yaml
@@ -556,7 +556,7 @@
 
 - group: system_info
   name: hw_revision
-  expert: False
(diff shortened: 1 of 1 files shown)


denniszollo:
Looks good to me.  Can you fill out the following truth table for me so I understand?

device          hw_ variant     hw_revision   product_id
bare multi:  
Duro:
Duro Inertial:

PR #870, opened 2018-09-07
Update INS settings documentation for console

denniszollo:


piksi_tools/console/settings.yaml
```diff
@@ -2199,3 +2199,202 @@
...
+- group: ins
+  name: output_mode
+  expert: false
+  type: enum
+  units: N/A
+  default value: 'disabled'
+  readonly: false
+  enumerated possible values: Disabled, Loosely Coupled, LC + GNSS
+  Description: Determines output mode of the inertial navigation outputs. 
+  Notes: Disabled configures device to output GNSS only solutions.
```
nicolekelley: Suggested comment:
```Disabled will disable the inertial solution - device is configured to output GNSS only solutions 
Loosely Coupled will output a loosely coupled GNSS and inertial solution
LC + GNSS configures a testing mode that simultaneously outputs both GNSS only and the loosely coupled inertial solution.
In LC + GNSS mode, there are two sets of solution messages with identical times. The flags indicate which messages are GNSS only solutions and which include are the loosely coupled inertial solution```

File: piksi_tools/console/settings.yaml [hunk 2]
```diff
@@ -2199,3 +2199,202 @@
...
+  Description: Noise parameter for odometry source 4
+
+- group: ins
+  name: vehicle_frame_roll
+  expert: false
+  type: double
+  units: degrees
+  default value: '0'
+  readonly: false
+  Description: Roll angle representing rotation from vehicle frame to device frame.
```
nicolekelley: I thought we were moving away from roll, pitch, and yaw and instead were going to describe the clockwise/counterclockwise rotations about an axix.

piksi_tools/console/settings.yaml
```diff
@@ -2199,3 +2199,202 @@
...
+               mounting bracket shipped with Duro is in use.
+
+- group: ins
+  name: vehicle_frame_offset_x
+  expert: false
+  type: double
+  units: meters
+  default value: '0'
+  readonly: false
+  Description: X component of vector from device frame to ins vehicle frame origin in which ins outputs are provided
```
nicolekelley: suggested change:
```The vector is measured in the device frame according to the markings on the device.  In order to output inertial solutions at the antenna phase center, this should be the same value (both sign and magnitude) as antenna_offset_x setting.```

File: piksi_tools/console/settings.yaml [hunk 4]
```diff
@@ -2199,3 +2199,202 @@
...
+- group: ins
+  name: vehicle_frame_offset_y
+  expert: false
+  type: double
+  units: meters
+  default value: '0'
+  readonly: false
+  Description: Y component of vector from device frame to ins vehicle frame origin in which ins outputs are provided  
+  Notes:       The vector is measured in the device frame according to the markings on the device.  In order to output
+               ins solutions at the antenna phase center, this should match in sign and magnitude of antenna_offset_y
```
nicolekelley: see above

piksi_tools/console/settings.yaml
```diff
@@ -2199,3 +2199,202 @@
...
+- group: ins
+  name: vehicle_frame_offset_z
+  expert: false
+  type: double
+  units: meters
+  default value: '-0.12674'
+  readonly: false
+  Description: Z component of vector from device frame to ins vehicle frame origin in which ins outputs are provided  
+  Notes:       The vector is measured in the device frame according to the markings on the device.  In order to output
+               ins solutions at the antenna phase center, this should match in sign and magnitude of antenna_offset_y
```
nicolekelley: see above

piksi_tools/console/settings.yaml
```diff
@@ -2199,3 +2199,202 @@
...
+  name: constrain_vehicle_sideslip
+  expert: true
+  type: boolean
+  units: N/A
+  default value: false
+  readonly: false
+  Description: Enables or disables experimental non-holonomic constraint feature that allows INS to make assumptions about vehicle dynamics 
+  Notes:       This settings should only be enabled provided the vehicle frame Eueler angles are measured precisely and are correct.
+               It assumes that a vehicle can have no body velocity in the direction aligned with the vehicle "y" axis 
+               (i.e no sideslip).  This is a reasonable assumption for passenger vehicles and many tractors.
```
nicolekelley: suggested description:
```Enables or disables non-holonomic constraint that allows the inertial solution to make assumptions about vehicle dynamics```

suggested notes:
```This settings should only be enabled provided the vehicle frame Euler angles are measured precisely and are correct. It assumes a vehicle can have no velocity in the direction aligned with the vehicle "y" axis (i.e no sideslip). This is a reasonable assumption for passenger vehicles and many tractors.```

Should this be moved to an advanced setting since this is a non-supported feature?

File: piksi_tools/console/settings.yaml [hunk 7]
```diff
@@ -2199,3 +2199,202 @@
...
+
+- group: ins
+  name: dr_duration_max
+  expert: true
+  type: double
+  units: seconds
+  default value: '10'
+  readonly: false
+  Description: Indicates the maximum duration in seconds for which the INS system will dead-reckon
```
nicolekelley: INS = inertial solution

piksi_tools/console/settings.yaml
```diff
@@ -2199,3 +2199,202 @@
...
+  Notes:       The default value of 10 seconds was chosen as a duration for which the Duro Inertial INS
```
nicolekelley: suggested change:
```The default value of 10 seconds was chosen as the expected duration for which the Duro Inertial solution can maintain sub-meter accuracy.```

File: piksi_tools/console/settings.yaml [hunk 9]
```diff
@@ -2199,3 +2199,202 @@
...
+  Description: ins build name
+
+- group: ins
+  name: build_date
+  expert: true
+  type: string
+  units: N/A
+  default value: N/A
+  readonly: true
+  Description: ins build date
```
nicolekelley: ins = Inertial

piksi_tools/console/settings.yaml
```diff
@@ -2199,3 +2199,202 @@
...
+               system can maintain sub-meter accuracy.
+
+- group: ins
+  name: build_name
+  expert: true
+  type: string
+  units: N/A
+  default value: N/A
+  readonly: true
+  Description: ins build name
```
nicolekelley: ins = Inertial

piksi_tools/console/settings.yaml
denniszollo: I will remove double spaces.  I will remove all abbreviations of INS in the descriptions.

piksi_tools/console/settings.yaml
denniszollo: I was hoping to do that, but I am just leaving well enough alone.  Forthcoming are user docs to explain these and provide a worksheet to configure.

piksi_tools/console/settings.yaml
denniszollo: it is already advanced

PR #883, opened 2018-09-20
Update fileio tool to allow specifing source and dest directories

denniszollo:


piksi_tools/fileio.py
```diff
@@ -365,14 +365,16 @@ def hexdump(data):
...
     while data:
+        # get 16 bytes from byte array and store in "chunk"
         chunk = data[:16]
+        # remove bytes from data
         data = data[16:]
         s = "%08X  " % ofs
-        s += " ".join("%02X" % ord(c) for c in chunk[:8]) + "  "
-        s += " ".join("%02X" % ord(c) for c in chunk[8:])
+        s += " ".join("%02X" % c for c in chunk[:8]) + "  "
+        s += " ".join("%02X" % c for c in chunk[8:])
```
silverjam: Can you explain what this is doing?

piksi_tools/fileio.py
```diff
@@ -457,13 +468,23 @@ def main():
             f = FileIO(link)
             try:
                 if args.write:
-                    f.write(args.write[0], open(args.write[0]).read())
+                    f.write(args.write[1], open(args.write[0]).read())
                 elif args.read:
+                    assert len(args.read) <= 2, "Only one source and one destination accepted for sbp fileio read."
```
silverjam: Might be nice to have something a little more user friendly:
```python
if len(args.read) not in [1, 2]:
  sys.stderr.write("ERROR: ...\n")
  sys.exit(1)
```

piksi_tools/fileio.py
```diff
@@ -457,13 +468,23 @@ def main():
...
+
                     data = f.read(args.read[0])
                     if args.hex:
-                        print(hexdump(data))
+                        if len(args.read) == 2:
+                            with open(args.read[1], 'w') as fd:
```
silverjam: You can invert this if to avoid some duplicate code (also probably want to open the file with `'wb'`):

```python
if len(args.read) == 2:
    with open(args.read[1], 'wb') as fd:
      fd.write(hexdump(data) if args.hex else data)
else:
  print(hexdump(data) if args.hex else data)
```

piksi_tools/fileio.py
denniszollo: This is creating a HEX dump a la a typical pc hex editor.  It puts a dump of HEX in columns on the left, followed by the PIPE character, followed by ASCII of the same bytes.  Bytes that can't be shown as ascii are shown as a period.

The input here changed from raw characters to a bytearray, so I had to update this code.

piksi_tools/fileio.py
denniszollo: Actually I need to open as 'w' since the hex dump is actually formatted text in this case.  I will take the cleaner flow control though thanks!

denniszollo: I take that back, I need to open as binary.

piksi_tools/fileio.py
denniszollo:
```00000000  70 69 70 3D 3D 31 38 2E  30 0A 6E 75 6D 70 79 3D  |pip==18.0.numpy=|
00000010  3D 31 2E 31 33 2E 31 0A  63 79 74 68 6F 6E 3D 3D  |=1.13.1.cython==|
00000020  30 2E 32 36 0A 70 79 74  65 73 74 3D 3D 33 2E 32  |0.26.pytest==3.2|
00000030  2E 31 0A 63 6F 6E 66 69  67 70 61 72 73 65 72 0A  |.1.configparser.|
00000040  74 6F 78 3D 3D 33 2E 32  2E 31 20 0A 66 75 74 75  |tox==3.2.1 .futu|```

denniszollo: left column is offset, then you have 16 bytes, then 16 characters as the snippet above exemplifies.

silverjam: Ok, cool, thanks for clarifying

PR #904, opened 2018-10-09
Fix upgrade fail for -dirty builds [DEVC-1178]

yizow:
parse_version strips "-dirty" instead of just "dirty". The dash caused
the version to be interpreted as "LegacyVersion", messing up version
comparison logic.

### DIFF
--- a/piksi_tools/console/update_view.py
+++ b/piksi_tools/console/update_view.py
@@ -54,7 +54,7 @@ def parse_version(version):
     if version[0] == 'v':
         version = version[1:]
     return pkparse_version(version.replace(
-        "dirty",
+        "-dirty",
         "", ))
 
 


yizow:
Tested by trying to upgrade from `2.1.6-dirty` (as reported by Swift Console) to `PiksiMulti-*-v2.1.6-dirty.bin` and `PiksiMulti-*-v2.1.6.bin`.
All can upgrade

Note: Modified upgrade function to only report whether it _would have_ upgraded or not, didn't actually perform upgrade

yizow:
Tested by trying to upgrade from `2.1.6-dirty` (as reported by Swift Console) to `PiksiMulti-*-v2.1.6-dirty.bin` and `PiksiMulti-*-v2.1.6.bin`.
All can upgrade

Note: Modified upgrade function to only report whether it _would have_ upgraded or not, didn't actually perform upgrade

PR #911, opened 2018-10-12
[MAP-696] Delete info pertaining to dynamics filter settings.

kmdade:


kmdade:
@lloydmaza I am going to delete these so that they don't accumulate dust in this repository. If in the end we need to have the same (or different) settings reinstated this is the place to look.  Note that this change doesn't affect what settings are in the console, just whether or not they have full support (description strings, etc.).

lloydmaza:
Solid - I will probably be revisiting this file sometime soon. 

Changes all LGTM.

PR #913, opened 2018-10-19
don't reuse last_soln for centering on view

denniszollo:


denniszollo:
I may still want to merge this.  TODO: test centering on Master

PR #914, opened 2018-10-24
Add more settings update error codes [DEVC-1225] [DEVC-1219]

yizow:
Recognize any positive sbp_msg.status as an error

See also:
https://github.com/swift-nav/piksi_buildroot/pull/958

### DIFF
--- a/piksi_tools/console/settings_view.py
+++ b/piksi_tools/console/settings_view.py
@@ -14,6 +14,7 @@ from __future__ import absolute_import, print_function
 import threading
 import time
 import configparser
+from enum import IntEnum, unique
 
 from pyface.api import GUI
 from sbp.piksi import MsgReset
@@ -46,6 +47,13 @@ if ETSConfig.toolkit != 'null':
 else:
     SVGButton = dict
 
+@unique
+class SBP_WRITE_STATUS(IntEnum):
+    TIMED_OUT           = -1
+    VALUE_REJECTED      = 1
+    SETTING_REJECTED    = 2
+    PARSE_FAILED        = 3
+    VALUE_READ_ONLY     = 4
 
 class TimedDelayStoppableThread(threading.Thread):
     """Thread class with a stop() method. The thread itself has to check
@@ -155,7 +163,7 @@ class Setting(SettingBase):
                 self.readonly = True
         self._prevent_revert_thread = False
 
-    def revert_to_prior_value(self, name, old, new, error_value=-1):
+    def revert_to_prior_value(self, name, old, new, error_value=SBP_WRITE_STATUS.TIMED_OUT):
         '''Revert setting to old value in the case we can't confirm new value'''
 
         if self.readonly:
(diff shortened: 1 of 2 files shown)


piksi_tools/console/settings_view.py
```diff
@@ -155,7 +155,7 @@ def __init__(self, name, section, value, ordering=0, settings=None):
                 self.readonly = True
         self._prevent_revert_thread = False
 
-    def revert_to_prior_value(self, name, old, new):
+    def revert_to_prior_value(self, name, old, new, error_value=1):
```
benjaminaltieri: Due to the behavior here:
https://github.com/swift-nav/piksi_tools/blob/1cbcd2c0e373b11aa1d2f0edfbede827895dd5e8/piksi_tools/console/settings_view.py#L211-L216
Let's create a timeout error condition and make that the default, which we can make -1 since that couldn't possibly come from an sbp message, which would use the `Unable to confirm that {0} was set to {1}.` verbiage. All other calls to this function should have an actual value from the device.

piksi_tools/console/settings_view.py
```diff
@@ -170,10 +170,33 @@ def revert_to_prior_value(self, name, old, new):
         invalid_setting_prompt = prompt.CallbackPrompt(
             title="Settings Write Error",
             actions=[prompt.close_button], )
-        invalid_setting_prompt.text = \
-            ("\n   Unable to confirm that {0} was set to {1}.\n"
-             "   Ensure the range and formatting of the entry are correct.\n"
-             "   Ensure that the new setting value did not interrupt console communication.").format(self.name, new)
+        if error_value == 1:    # SBP_WRITE_STATUS_VALUE_REJECTED
```
benjaminaltieri: Let's define these values near the top of the module somewhere (i.e. try to reduce magic numbers)

piksi_tools/console/settings_view.py
```diff
@@ -170,10 +170,33 @@ def revert_to_prior_value(self, name, old, new):
...
+            invalid_setting_prompt.text = \
+                ("\n   Unable to confirm that {0} was set to {1}.\n"
+                 "   Ensure the range and formatting of the entry are correct.\n"
+                 "   Ensure that the new setting value did not interrupt console communication.\n"
+                 "   Error Value: {2}")
+        elif error_value == 2:  # SBP_WRITE_STATUS_SETTING_REJECTED
+            invalid_setting_prompt.text = \
+                ("\n   Unable to confirm that {0} was set to {1}.\n"
```
benjaminaltieri: We can replace this with something along the lines of "Setting not registered in device, refresh settings to see currently available settings"

piksi_tools/console/settings_view.py
```diff
@@ -170,10 +170,33 @@ def revert_to_prior_value(self, name, old, new):
...
+                 "   Unknown setting.\n"
+                 "   Error Value: {2}")
+        elif error_value == 3:  # SBP_WRITE_STATUS_PARSE_FAILED
+            invalid_setting_prompt.text = \
+                ("\n   Unable to confirm that {0} was set to {1}.\n"
+                 "   Value not parseable.\n"
+                 "   Error Value: {2}")
+        elif error_value == 4:  # SBP_WRITE_STATUS_VALUE_READ_ONLY
+            invalid_setting_prompt.text = \
+                ("\n   Unable to confirm that {0} was set to {1}.\n"
```
benjaminaltieri: Replace "Unable to confirm" wording with "Unable to set x to y"

piksi_tools/console/settings_view.py
```diff
@@ -170,10 +170,33 @@ def revert_to_prior_value(self, name, old, new):
...
+                 "   Setting is read-only.\n"
+                 "   Error Value: {2}")
+        else:
```
benjaminaltieri: Add case for timeout.

benjaminaltieri:
Looks like Travis is having an issue here, do we need to modify the install requirements?
`ImportError: No module named enum`

benjaminaltieri:
Probably just add `enum32` to `requirement.txt`. I would think later versions should be compatible with having it explicitly installed.

benjaminaltieri:
Did a quick `pip install enum32` and `pip freeze`, you should be able to pin to `enum34==1.1.6`

PR #916, opened 2018-10-30
Make sure we get the size of the file correct

silverjam:
The current code has an issue where the file will be padded out
with NULL bytes once we've started reading past the end of the file.

Instead of trying to compute the exact size of the file buffer, just
record the data chunks and the offsets they correspond to.

### DIFF
--- a/piksi_tools/fileio.py
+++ b/piksi_tools/fileio.py
@@ -12,6 +12,7 @@
 from __future__ import absolute_import, print_function
 
 import copy
+import itertools
 import random
 import time
 import threading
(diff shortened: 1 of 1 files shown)


denniszollo:
Does this only affect the "read" call ?

piksi_tools/fileio.py
```diff
@@ -238,27 +239,31 @@ def read(self, filename):
...
+        closure = {'mostly_done': False, 'done': False, 'buf': {}, 'pending':set()}
 
         def cb(req, resp):
-            if len(closure['buf']) < req.offset:
-                closure['buf'] += [0] * (req.offset - len(closure['buf']))
-            closure['buf'][req.offset:
-                           req.offset + len(resp.contents)] = resp.contents
+            closure['pending'].remove(req.offset)
+            closure['buf'][req.offset] = resp.contents
             if req.chunk_size != len(resp.contents):
```
denniszollo: Is this condition guarenteed to happen ONLY on the last packet of the file?

denniszollo: By "this" I mean when the chunk_size isn't same as the response contents
```if req.chunk_size != len(resp.contents):```

silverjam: If we ask for ~250 bytes and get back 0 or <250 for the `open()`/`read()` calls on Linux you're guaranteed that something went "wrong", either the call failed, or you reached the end of the file.  If you read past the end of the file in the fileio daemon we just return an empty buffer because the `seek()` call on the file will have failed.  This always seems to be how we detected the end of the file, the new behavior is the `fileio` module having multiple read packets in-flight through the SelectiveRepeater, so we can potentially attempt to read `WINDOW_SIZE*MAX_PACKET_SIZE` bytes past the end of the file.

PR #917, opened 2018-11-13
Create tcp driver util and wrap error handling [ESD-632]

benjaminaltieri:
Requires https://github.com/swift-nav/libsbp/pull/617 in order to work.

This change will wrap all attempts at TCP connection using the `libsbp` driver with standard error codes and a parsing scheme that seems to be common amongst tools.

Introduces the behavior that a timeout will cause the initialization to fail rather than hang.

### DIFF
--- a/piksi_tools/fileio.py
+++ b/piksi_tools/fileio.py
@@ -19,7 +19,7 @@ import threading
 import sys
 
 from sbp.client import Framer, Handler
-from sbp.client.drivers.network_drivers import TCPDriver
+from piksi_tools.utils import get_tcp_driver
 from sbp.file_io import (SBP_MSG_FILEIO_READ_DIR_RESP,
                          SBP_MSG_FILEIO_READ_RESP, SBP_MSG_FILEIO_WRITE_RESP,
                          MsgFileioReadDirReq, MsgFileioReadDirResp,
@@ -459,11 +459,7 @@ def main():
     port = args.port[0]
     baud = args.baud[0]
     if args.tcp:
-        try:
-            host, port = port.split(':')
-            selected_driver = TCPDriver(host, int(port))
-        except:  # noqa
-            raise Exception('Invalid host and/or port')
+        selected_driver = get_tcp_driver(port)
     else:
         selected_driver = serial_link.get_driver(args.ftdi, port, baud)
 
--- a/piksi_tools/fileio.py
+++ b/piksi_tools/fileio.py
@@ -19,13 +19,13 @@ import threading
 import sys
 
 from sbp.client import Framer, Handler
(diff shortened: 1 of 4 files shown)


piksi_tools/utils.py
```diff
@@ -38,3 +41,21 @@ def sopen(path, mode):
...
+        return TCPDriver(host,
+                         int(port),
+                         raise_initial_timeout=True)
+    except ValueError:
+        raise Exception('Invalid format (use ip_address:port): {}'.format(host))
+    except socket.timeout:
+        raise Exception('TCP connection timed out. Check host: {}'.format(host))
+    except Exception as e:
+        import traceback
+        raise Exception('Invalid host and/or port: {0}'.format(traceback.format_exc()))
```
silverjam: Will this be displayed to a user?  Is that really what we want?  Seems like we could at least try to catch some more common errors before displaying a traceback?

benjaminaltieri: I believe the current behavior is that the console merely exits, without any dialogue at all. I definitely see your point though, and admittedly, I was keeping this scope relatively focused on avoiding the spinning pinwheel (silent hanging of the connection), rather than refactor the entire feedback cycle.

yizow: In terms of trying to connect:
1) Invalid host:port format
2) host unreachable
seem to be the most common errors; ValueError and socket.timeout catch them, respectively.

denniszollo: I have the feedback method of popping up the "help" dialog when people incorrectly format command line options.   Will this dialog show up in the case of bad / missing host and port?

yizow: Yes: if command line options are incorrect, the help dialog will pop up before any of this new code gets reached.

piksi_tools/serial_link.py
```diff
@@ -31,7 +31,7 @@
 from sbp.logging import SBP_MSG_LOG, SBP_MSG_PRINT_DEP, MsgLog
 from sbp.piksi import MsgReset
 
-from piksi_tools.utils import mkdir_p
+from piksi_tools.utils import (mkdir_p, get_tcp_driver)
```
yizow: is there a reason for the parentheses? Only asking because inconsistent style with above

piksi_tools/fileio.py
```diff
@@ -19,7 +19,7 @@
 import sys
 
 from sbp.client import Framer, Handler
-from sbp.client.drivers.network_drivers import TCPDriver
+from piksi_tools.utils import get_tcp_driver
```
yizow: maybe change order of imports so all `sbp.*` are imported together

PR #925, opened 2018-11-16
Create tcp driver util and wrap error handling [ESD-632]

benjaminaltieri:
release version of #917


denniszollo:
looks good, long term I would love if all the drivers had a unified API so we could respond when the interface disappears or times out in a unified fashion.

PR #926, opened 2018-11-16
Allow tracking view to age out old data [ESD-902]

silverjam:
Tracking view adds data to it's `CN0_dict` memory but never ages out old data, thus anything that shows up will forever be held in the dictionary, this meant that logic to delete plots and plot data was never triggered.  Add a mechanism to figure out if data in the dictionary is older than the oldest currently active plot point and remove it if it is.

### DIFF
--- a/piksi_tools/console/tracking_view.py
+++ b/piksi_tools/console/tracking_view.py
@@ -10,6 +10,7 @@
 # EITHER EXPRESSED OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE IMPLIED
 # WARRANTIES OF MERCHANTABILITY AND/OR FITNESS FOR A PARTICULAR PURPOSE.
 
+import sys
 import time
 from collections import defaultdict, deque
 import threading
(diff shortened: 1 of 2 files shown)


piksi_tools/console/tracking_view.py
```diff
@@ -133,104 +133,123 @@ class TrackingView(CodeFiltered):
...
+            for k, cno_array in self.CN0_dict.items():
+                key = str(k)
+                # set plot data
+                if (getattr(self, 'show_{}'.format(int(k[0])), True)):
+                    new_plot_data[key] = cno_array
+            self.plot_data.update_data(new_plot_data)
+            # Remove any stale plots that got removed from the dictionary
+            for each in list(self.plot.plots.keys()):
+                if each not in [str(a) for a in self.CN0_dict.keys()] and each != 't':
+                    try:
```
silverjam: This is a section of code that was never triggered because `CN0_dict` retains everything forever

piksi_tools/console/tracking_view.py
```diff
@@ -133,104 +133,123 @@ class TrackingView(CodeFiltered):
...
-                    pass
-        # add/remove plot as neccesary and build legend
-        for k, cno_array in self.CN0_dict.items():
-            key = str(k)
-            if (getattr(self, 'show_{}'.format(int(k[0])), True) and
-               not cno_array.count(0) == NUM_POINTS):
-                if key not in self.plot.plots.keys():
-                    pl = self.plot.plot(('t', key), type='line',
-                                        color=get_color(k), name=key)
+        with self.CN0_lock:
```
silverjam: Use a `with` to we can have an early return and still drop the lock easily.

piksi_tools/console/tracking_view.py
```diff
@@ -133,104 +133,123 @@ class TrackingView(CodeFiltered):
...
-                self.CN0_dict[key].append(s.cn0 / 4.0)
-            received_code_list = getattr(self, "received_codes", [])
-            if s.sid.code not in received_code_list:
-                received_code_list.append(s.sid.code)
-                self.received_codes = received_code_list
-        for key, cno_array in self.CN0_dict.items():
-            if key not in codes_that_came:
-                cno_array.append(0)
-        self.CN0_lock.release()
+        with self.CN0_lock:
```
silverjam: Use a `with` to we can have an early return and still drop the lock easily.

piksi_tools/console/tracking_view.py
```diff
@@ -133,104 +133,123 @@ class TrackingView(CodeFiltered):
...
+                    self.CN0_age[key] = t
+                received_code_list = getattr(self, "received_codes", [])
+                if s.sid.code not in received_code_list:
+                    received_code_list.append(s.sid.code)
+                    self.received_codes = received_code_list
+            for key, cno_array in list(self.CN0_dict.items()):
+                if key not in codes_that_came:
+                    cno_array.append(0)
+            self.clean_cn0(t)
+            if self.update_scheduled:
```
silverjam: Avoid scheduling an update if one is already in-flight, this should reduce some of the processing overhead for plotting.

piksi_tools/console/tracking_view.py
```diff
@@ -133,104 +133,123 @@ class TrackingView(CodeFiltered):
...
+                if code_is_glo(s.sid.code):
+                    sat = s.fcn - GLO_FCN_OFFSET
+                    self.glo_slot_dict[i] = s.sid.sat
+                else:
+                    sat = s.sid.sat
+                key = (s.sid.code, sat, i)
+                codes_that_came.append(key)
+                if s.cn0 != 0:
+                    self.CN0_dict[key].append(s.cn0 / 4.0)
+                    self.CN0_age[key] = t
```
silverjam: Update the "age" of the data in the `CN0_dict`

piksi_tools/console/tracking_view.py
```diff
@@ -133,104 +133,123 @@ class TrackingView(CodeFiltered):
...
-        for key, cno_array in self.CN0_dict.items():
-            if key not in codes_that_came:
-                cno_array.append(0)
-        self.CN0_lock.release()
+                    sat = s.mesid.sat
+                key = (s.mesid.code, sat, i)
+                codes_that_came.append(key)
+                if s.cn0 != 0:
+                    self.CN0_dict[key].append(s.cn0 / 4.0)
+                    self.CN0_age[key] = t
```
silverjam: Update the "age" of the data in the `CN0_dict`

piksi_tools/console/tracking_view.py
```diff
@@ -133,104 +133,123 @@ class TrackingView(CodeFiltered):
                 Item('legend_visible', label="Show Legend:"),
                 CodeFiltered.get_filter_group(), )))
 
+    def clean_cn0(self, t):
+        assert self.CN0_lock.locked()
+        for k in list(self.CN0_dict.keys()):
+            if self.CN0_age[k] < self.time[0]:
```
silverjam: If the age of the data in the CN0 dict is older than the "oldest active time" then delete the data so it can be cleared out of the plots and the plot data.

silverjam:
@denniszollo Please see GitHub comments inline for call-outs of what changed (with explanations)-- the introduction of the `with` statement obscured a lot of the changes.

piksi_tools/console/tracking_view.py
```diff
@@ -133,104 +133,123 @@ class TrackingView(CodeFiltered):
                 Item('legend_visible', label="Show Legend:"),
                 CodeFiltered.get_filter_group(), )))
 
+    def clean_cn0(self, t):
+        assert self.CN0_lock.locked()
+        for k in list(self.CN0_dict.keys()):
```
yizow: `list()` shouldn't be necessary here, `.keys()` returns a copy
https://docs.python.org/2.7/library/stdtypes.html#dict.keys

piksi_tools/console/tracking_view.py
```diff
@@ -133,104 +133,123 @@ class TrackingView(CodeFiltered):
...
+            # Update the underlying plot data from the CN0_dict for selected items
+            new_plot_data = {'t': self.time}
+            for k, cno_array in self.CN0_dict.items():
+                key = str(k)
+                # set plot data
+                if (getattr(self, 'show_{}'.format(int(k[0])), True)):
+                    new_plot_data[key] = cno_array
+            self.plot_data.update_data(new_plot_data)
+            # Remove any stale plots that got removed from the dictionary
+            for each in list(self.plot.plots.keys()):
```
yizow: as above, shouldn't need `list()`

piksi_tools/console/tracking_view.py
```diff
@@ -133,104 +133,123 @@ class TrackingView(CodeFiltered):
...
-                plots.append(pl)
-                plot_labels.append(get_label(k, self.glo_slot_dict))
-            # if not selected or all 0, remove
-            else:
-                if key in self.plot.plots.keys():
-                    self.plot.delplot(key)
-        plots = dict(zip(plot_labels, plots))
-        self.plot.legend.plots = plots
-        self.CN0_lock.release()
+                    if key in list(self.plot.plots.keys()):
```
yizow: as above, `list()` not necessary

piksi_tools/console/tracking_view.py
silverjam: It's necessary here since CN0_dict is being modified inside the iterator

piksi_tools/console/tracking_view.py
silverjam: This was initially from a Python 3 compatibility tool (`2to3`) but just in case `self.plot.delplot(each)` modifies the underlying data structure behind `list(self.plot.plots.keys())` I'd like to retain the `list` call.

piksi_tools/console/tracking_view.py
yizow: ah, my bad; i thought this was python 2

silverjam: Well, it is and it isn't, I think this backward (and forward) compatible with both.  Going forward we're going to want Python 3 support, so I think it's safe to introduce some compat stuff now.

silverjam:
@dzollo Feedback?

PR #927, opened 2018-11-21
EX-445: Add rtcm_out expert parameters

nsirola:
Add the settings for antenna height and antenna/receiver descriptors from https://github.com/swift-nav/piksi_buildroot/pull/1010. See issue https://swift-nav.atlassian.net/browse/EX-445


nsirola:
Satisfied flake8 by adding the whitespace where it wanted. Strange how the previous PRs went through just fine...

PR #934, opened 2018-11-27
remove channel from plotting key [ESD-614]

yizow:
Channel is entirely local, based only on the order in which we receive
satellite measurements. If the piksi changes this order (for example, if
it lost and then reacquired signals), then the channel will likely
change, rendering it useless as an identifying metric.

### DIFF
--- a/piksi_tools/console/tracking_view.py
+++ b/piksi_tools/console/tracking_view.py
@@ -77,7 +77,7 @@ color_dict = {
 
 
 def get_color(key):
-    code, sat, ch = key
+    code, sat = key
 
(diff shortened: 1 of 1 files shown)


silverjam:
Looks reasonable, what's remaining to remove the `{WIP}` tag?

benjaminaltieri:
looks good to me too!

yizow:
Oops, forgot to remove {WIP}

silverjam:
🚢 🇮🇹

silverjam:
@yizow backport?

PR #937, opened 2018-11-29
Refactor & tweak status bar mode determination [ESD-807] [ESD-841]

tn-exafore:
There are / were two related bugs in status bar mode in ESD-807:
1. Flicker between SBAS / Fixed RTK, observed by Swift (the original bug)
2. Flicker between None / Fixed RTK, observed by Exafore

This PR only fixes the latter one as we were unable to reproduce the previous one. Possibly it can also fix the previous one but that is uncertain, to be found out in testing.

The fix is made by reverting the timing aspect of ESD-697 (302b252f8b5fa). Otherwise the idea of that task is kept - show mode from baseline tab if that is in differential mode but llh/solution isn't, and mode from solution tab otherwise. But some refactoring is done to improve code readability. Frankly, to me the exact best/wanted solution functionality-wise is a bit unclear but this is what I deduced from task comments in ESD-807 and ESD-697, and the commit 302b252f8b5fa.

### DIFF
--- a/piksi_tools/console/console.py
+++ b/piksi_tools/console/console.py
@@ -56,7 +56,7 @@ from piksi_tools.console.system_monitor_view import SystemMonitorView
 from piksi_tools.console.tracking_view import TrackingView
 from piksi_tools.console.update_view import UpdateView
 from piksi_tools.console.utils import (EMPTY_STR, call_repeatedly,
-                                       get_mode, mode_dict, resource_filename,
+                                       mode_dict, resource_filename,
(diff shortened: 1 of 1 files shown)


tn-exafore:
created https://github.com/swift-nav/piksi_tools/pull/945 to do the backporting to v2.2.0

tn-exafore:
@pmiettinen please re-review after some tweaks

pmiettinen:
LGTM

PR #938, opened 2018-11-29
Use libsettings [ESD-873]

pmiettinen:
https://github.com/swift-nav/libsettings/pull/25

# Testing
## Currently known defects with libsettings-0.1.9
- [x] `libsettings` `signal` is triggered before `wait` is finished resulting to exception where the event object is `None`. This has been irregularly seen on console startup when using TCP connection. `libsettings` offers a mechanism to do this properly via `wait_init` where event can be made guaranteed to exist before the request is sent and response received.
FYI: One can revive from this state by pushing the `Refresh settings` button.

## Currently known defects with libsettings-0.1.10
- [x] Cython exports have strings incompatible with Python3
- [x] Performance drop in read/write all cases

## TODO
- [x] `pip install -e ./`
  - [x] tag libsettings-v0.1.11
  - [x] push it to PyPI after we are convinced it's stable/working
  - [x] update requirements.txt
- [x] Suppress unnecessary warning spam in case of device/connection failure during write/read all

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -12,6 +12,11 @@ environment:
       PYTHON_VERSION: 3.5
       PYTHON_ARCH: 32
 
+  APPVEYOR_RDP_PASSWORD: Libsettings911!
+
+init:
+  - ps: iex ((new-object net.webclient).DownloadString('https://raw.githubusercontent.com/appveyor/ci/master/scripts/enable-rdp.ps1'))
+
 
 cache:
   # Cache downloaded pip packages and built wheels.
@@ -27,10 +32,22 @@ install:
   - cinst nsis
   - SET PATH=%PYTHON%;%PYTHON%\Scripts;%PATH%
   - SET PYTHONPATH=.
+  - set PATH=C:\MinGW\bin;%PATH%
   - python --version
   # Upgrade to the latest pip.
   - '%CMD_IN_ENV% python -m pip install -U pip setuptools wheel tox'
   - powershell .\\tasks\\missing-headers.ps1
+  - md lib
+  - cd lib
+  - appveyor DownloadFile "https://raw.githubusercontent.com/swift-nav/libsettings/v0.1.3/python/libsettings-0.1.3.tar.gz"
+  - md libsettings
+  - tar -xzf libsettings-0.1.3.tar.gz -C libsettings --strip-components=1
+  - cd libsettings
+  # Hack to get rid of format specifier warnings emerging from the cythonized file
+  - C:\cygwin64\\bin\sed.exe -i 's/#define CYTHON_FORMAT_SSIZE_T "z"/#define CYTHON_FORMAT_SSIZE_T "I"/g' python/libsettings.c
+  - C:\cygwin64\\bin\sed.exe -i "s/zd /Id /g" python/libsettings.c
+  - python setup.py build_ext --inplace --compiler=mingw32
+  - cd ../..
   - '%CMD_IN_ENV% python scripts\build_release.py'
 
 build: off
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -12,11 +12,6 @@ environment:
       PYTHON_VERSION: 3.5
       PYTHON_ARCH: 32
 
-  APPVEYOR_RDP_PASSWORD: Libsettings911!
-
-init:
-  - ps: iex ((new-object net.webclient).DownloadString('https://raw.githubusercontent.com/appveyor/ci/master/scripts/enable-rdp.ps1'))
-
 
 cache:
   # Cache downloaded pip packages and built wheels.
@@ -32,22 +27,10 @@ install:
   - cinst nsis
   - SET PATH=%PYTHON%;%PYTHON%\Scripts;%PATH%
   - SET PYTHONPATH=.
-  - set PATH=C:\MinGW\bin;%PATH%
   - python --version
   # Upgrade to the latest pip.
   - '%CMD_IN_ENV% python -m pip install -U pip setuptools wheel tox'
   - powershell .\\tasks\\missing-headers.ps1
-  - md lib
-  - cd lib
-  - appveyor DownloadFile "https://raw.githubusercontent.com/swift-nav/libsettings/v0.1.3/python/libsettings-0.1.3.tar.gz"
-  - md libsettings
-  - tar -xzf libsettings-0.1.3.tar.gz -C libsettings --strip-components=1
-  - cd libsettings
-  # Hack to get rid of format specifier warnings emerging from the cythonized file
-  - C:\cygwin64\\bin\sed.exe -i 's/#define CYTHON_FORMAT_SSIZE_T "z"/#define CYTHON_FORMAT_SSIZE_T "I"/g' python/libsettings.c
-  - C:\cygwin64\\bin\sed.exe -i "s/zd /Id /g" python/libsettings.c
-  - python setup.py build_ext --inplace --compiler=mingw32
-  - cd ../..
   - '%CMD_IN_ENV% python scripts\build_release.py'
 
 build: off
--- a/piksi_tools/console/console.py
+++ b/piksi_tools/console/console.py
@@ -145,10 +145,6 @@ class SwiftConsole(HasTraits):
       Update the firmware
     log_level_filter : str
       Syslog string, one of "ERROR", "WARNING", "INFO", "DEBUG".
-    skip_settings : bool
-      Don't read the device settings. Set to False when the console is reading
-      from a network connection only.
-
     """
 
     link = Instance(sbpc.Handler)
@@ -586,7 +582,6 @@ class SwiftConsole(HasTraits):
                  link,
                  update,
                  log_level_filter,
-                 skip_settings=False,
                  error=False,
                  cnx_desc=None,
                  json_logging=False,
@@ -715,11 +710,16 @@ class SwiftConsole(HasTraits):
                 if self.networking_view.connect_when_uuid_received:
                     self.networking_view._connect_rover_fired()
 
+            skip_settings_read = False
+            if 'mode' in self.connection_info:
+                if self.connection_info['mode'] == 'file':
+                    skip_settings_read = True
+
             settings_read_finished_functions.append(update_serial)
             self.settings_view = SettingsView(
                 self.link,
                 settings_read_finished_functions,
-                skip=skip_settings)
+                skip_read=skip_settings_read)
             self.update_view.settings = self.settings_view.settings
             self.python_console_env = {
                 'send_message': self.link,
--- a/piksi_tools/console/settings_view.py
+++ b/piksi_tools/console/settings_view.py
@@ -58,10 +58,8 @@ class WorkQueue():
     def _work_thd(self):
         while True:
             (func, argv) = self._work_queue.get(block=True)
-            self._settings_view.disable()
             func(*argv)
             self._work_queue.task_done()
-            self._settings_view.enable()
 
 
 class SettingBase(HasTraits):
@@ -329,7 +327,6 @@ class SettingsView(HasTraits):
       use when reading from network connections or file.
     """
     show_auto_survey = Bool(False)
-    enable_buttons = Bool(True)
     settings_yaml = list()
     auto_survey = SVGButton(
         label='Auto\nSurvey',
@@ -390,7 +387,6 @@ class SettingsView(HasTraits):
                         'auto_survey',
                         show_label=False,
                         visible_when='show_auto_survey'),
-                    enabled_when='enable_buttons',
                 ),
                 HGroup(
                     Item('settings_read_button', show_label=False,
@@ -398,7 +394,6 @@ class SettingsView(HasTraits):
                     Item('', label="Refresh settings\nfrom device", padding=0),
                     Item('expert', show_label=False),
                     Item('', label="Show Advanced\nSettings", padding=0),
-                    enabled_when='enable_buttons',
                 ),
                 Item('selected_setting', style='custom', show_label=False),
             ),
@@ -758,12 +753,6 @@ class SettingsView(HasTraits):
 
     def __exit__(self, *args):
         self.cleanup()
-
-    def disable(self):
-        self.enable_buttons = False
-
-    def enable(self):
-        self.enable_buttons = True
 
     def __init__(self,
                  link,
--- a/piksi_tools/console/settings_view.py
+++ b/piksi_tools/console/settings_view.py
@@ -45,7 +45,8 @@ else:
 
 class WorkQueue():
 
-    def __init__(self):
+    def __init__(self, settings_view):
+        self._settings_view = settings_view
         self._work_queue = Queue()
         self._worker = threading.Thread(target=self._work_thd)
         self._worker.daemon = True
@@ -57,8 +58,10 @@ class WorkQueue():
     def _work_thd(self):
         while True:
             (func, argv) = self._work_queue.get(block=True)
+            self._settings_view.disable()
             func(*argv)
             self._work_queue.task_done()
+            self._settings_view.enable()
 
 
 class SettingBase(HasTraits):
@@ -327,6 +330,7 @@ class SettingsView(HasTraits):
       use when reading from network connections or file.
     """
     show_auto_survey = Bool(False)
+    enable_buttons = Bool(True)
     settings_yaml = list()
     auto_survey = SVGButton(
         label='Auto\nSurvey',
@@ -387,6 +391,7 @@ class SettingsView(HasTraits):
                         'auto_survey',
                         show_label=False,
                         visible_when='show_auto_survey'),
+                    enabled_when='enable_buttons',
                 ),
                 HGroup(
                     Item('settings_read_button', show_label=False,
@@ -394,6 +399,7 @@ class SettingsView(HasTraits):
                     Item('', label="Refresh settings\nfrom device", padding=0),
                     Item('expert', show_label=False),
                     Item('', label="Show Advanced\nSettings", padding=0),
+                    enabled_when='enable_buttons',
                 ),
                 Item('selected_setting', style='custom', show_label=False),
             ),
@@ -495,6 +501,8 @@ class SettingsView(HasTraits):
         confirm_prompt2.run(block=False)
 
     def _settings_unconfirm_all(self):
+        # Clear the tabular editor
+        del self.settings_list[:]
         for section in self.settings.keys():
             for name in self.settings[section].keys():
                 self.settings[section][name].confirmed_set = False
@@ -752,6 +760,12 @@ class SettingsView(HasTraits):
     def __exit__(self, *args):
         self.cleanup()
 
+    def disable(self):
+        self.enable_buttons = False
+
+    def enable(self):
+        self.enable_buttons = True
+
     def __init__(self,
                  link,
                  read_finished_functions=[],
@@ -761,9 +775,8 @@ class SettingsView(HasTraits):
                  skip_read=False):
         super(SettingsView, self).__init__()
         self.settings_api = Settings(link)
-        self.workqueue = WorkQueue()
+        self.workqueue = WorkQueue(self)
         self.expert = expert
-        self.show_auto_survey = False
         self.gui_mode = gui_mode
         self.settings = {}
         self.link = link
--- a/piksi_tools/console/settings_view.py
+++ b/piksi_tools/console/settings_view.py
@@ -203,7 +203,6 @@ class Setting(SettingBase):
 
     def _write_value(self, old, new):
         if (old != new and old is not Undefined and new is not Undefined):
-            self.value = self.value.encode('ascii', 'replace')
             self.confirmed_set = False
             res = self.settings.settings_api.write(self.section, self.name, new)
             if res == SettingsWriteResponseCodes.SETTINGS_WR_OK:
--- a/piksi_tools/console/settings_view.py
+++ b/piksi_tools/console/settings_view.py
@@ -15,6 +15,8 @@ import threading
 import time
 import configparser
 
+from queue import Queue
+
 from pyface.api import GUI
 from sbp.piksi import MsgReset
 from sbp.settings import MsgSettingsSave
@@ -28,7 +30,6 @@ from traitsui.tabular_adapter import TabularAdapter
 import piksi_tools.console.callback_prompt as prompt
 from piksi_tools.console.gui_utils import MultilineTextEditor
 from piksi_tools.console.utils import swift_path
-from piksi_tools.settings import KEY_ENCODING, VALUE_ENCODING
 from pyface.api import FileDialog, OK
 
 from .settings_list import SettingsList
@@ -40,8 +41,6 @@ if ETSConfig.toolkit != 'null':
     from enable.savage.trait_defs.ui.svg_button import SVGButton
 else:
     SVGButton = dict
-
-from queue import Queue
 
 
 class WorkQueue():
@@ -201,8 +200,7 @@ class Setting(SettingBase):
 
     def _write_value(self, old, new):
         if (old != new and old is not Undefined and new is not Undefined):
-            if type(self.value) == unicode:
-                self.value = self.value.encode('ascii', 'replace')
+            self.value = self.value.encode('ascii', 'replace')
             self.confirmed_set = False
             res = self.settings.settings_api.write(self.section, self.name, new)
             if res == SettingsWriteResponseCodes.SETTINGS_WR_OK:
--- a/piksi_tools/console/settings_view.py
+++ b/piksi_tools/console/settings_view.py
@@ -306,9 +306,9 @@ class SettingsView(HasTraits):
       Show expert settings (defaults to False)
     gui_mode : bool
       ??? (defaults to True)
-    skip : bool
+    skip_read : bool
       Skip reading of the settings (defaults to False). Intended for
-      use when reading from network connections.
+      use when reading from network connections or file.
     """
     show_auto_survey = Bool(False)
     settings_yaml = list()
@@ -742,7 +742,7 @@ class SettingsView(HasTraits):
                  name_of_yaml_file="settings.yaml",
                  expert=False,
                  gui_mode=True,
-                 skip=False):
+                 skip_read=False):
         super(SettingsView, self).__init__()
         self.settings_api = Settings(link)
         self.workqueue = WorkQueue()
@@ -758,8 +758,7 @@ class SettingsView(HasTraits):
         # No support for arguments currently.
         self.read_finished_functions = read_finished_functions
         self.setting_detail = SettingBase()
-        self.pending_settings = []
-        if not skip:
+        if not skip_read:
             try:
                 self._settings_read_all()
             except IOError:
--- a/piksi_tools/console/settings_view.py
+++ b/piksi_tools/console/settings_view.py
@@ -145,7 +145,7 @@ class Setting(SettingBase):
         if readonly:
             self.readonly = True
 
-    def revert_to_prior_value(self, name, old, new, error_value=SettingsWriteResponseCodes.SETTINGS_WR_TIMEOUT):
+    def revert_to_prior_value(self, section, name, old, new, error_value):
         '''Revert setting to old value in the case we can't confirm new value'''
 
         if self.readonly:
@@ -155,7 +155,7 @@ class Setting(SettingBase):
         self.value = old
         self.write_failure = True
         invalid_setting_prompt = prompt.CallbackPrompt(
-            title="Settings Write Error",
+            title="Settings Write Error: {}.{}".format(section, name),
             actions=[prompt.close_button], )
         if error_value == SettingsWriteResponseCodes.SETTINGS_WR_TIMEOUT:
             invalid_setting_prompt.text = \
@@ -163,11 +163,7 @@ class Setting(SettingBase):
                  "   Message timed out.\n"
                  "   Ensure that the new setting value did not interrupt console communication.\n"
                  "   Error Value: {2}")
-        else:
-            invalid_setting_prompt.text = \
-                ("\n   Unable to set {0} to {1}.\n")
-
-        if error_value == SettingsWriteResponseCodes.SETTINGS_WR_VALUE_REJECTED:
+        elif error_value == SettingsWriteResponseCodes.SETTINGS_WR_VALUE_REJECTED:
             invalid_setting_prompt.text += \
                 ("   Ensure the range and formatting of the entry are correct.\n"
                  "   Error Value: {2}")
@@ -200,13 +196,13 @@ class Setting(SettingBase):
         self.reverting = False
 
     def _write_value(self, old, new):
-        if (old != new and old is not Undefined and new is not Undefined):
+        if (old is not Undefined and new is not Undefined):
             self.confirmed_set = False
-            res = self.settings.settings_api.write(self.section, self.name, new)
-            if res == SettingsWriteResponseCodes.SETTINGS_WR_OK:
+            (error, section, name, value) = self.settings.settings_api.write(self.section, self.name, new)
+            if error == SettingsWriteResponseCodes.SETTINGS_WR_OK:
                 self.value = new
             else:
-                self.revert_to_prior_value(self.name, old, new, res)
+                self.revert_to_prior_value(self.section, self.name, old, new, res)
 
             self.confirmed_set = True
 
@@ -607,8 +603,8 @@ class SettingsView(HasTraits):
         except IOError as e:
             print('Unable to export settings to file due to IOError: {}'.format(e))
 
-    def _import_success(self):
-        print("Successfully imported settings from file.")
+    def _import_success(self, count):
+        print("Successfully imported {} settings from file.".format(count))
         confirm_prompt = prompt.CallbackPrompt(
             title="Save to device flash?",
             actions=[prompt.close_button, prompt.ok_button],
@@ -616,6 +612,14 @@ class SettingsView(HasTraits):
         confirm_prompt.text = "\n" \
             "  Settings import from file complete.  Click OK to save the settings  \n" \
             "  to the device's persistent storage.  \n"
+        confirm_prompt.run(block=False)
+
+    def _import_fail(self):
+        confirm_prompt = prompt.CallbackPrompt(
+            title="Failed to import settings from file",
+            actions=[prompt.close_button])
+        confirm_prompt.text = "\n" \
+            "  Verify what config file is not overwriting active connection settings.  \n"
         confirm_prompt.run(block=False)
 
     def _import_failure_section(self, section):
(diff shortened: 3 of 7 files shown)


piksi_tools/console/settings.yaml
```diff
@@ -2242,7 +2242,7 @@
   readonly: false
   units: N/A
   Description: Enables or disables the Over-The-Air upgrade daemon's verbose output.
-  Notes: 'ota enabled' must be 'False' in order to change this setting.
+  Notes: The OTA daemon must be disabled in order to change this setting.
```
pmiettinen: My yaml parser didn't like the `'`s

pmiettinen:
> what is the plan for piksi_tools/settings.py ?

Can/Shall be updated in similar fashion to `settings_view.py`.

piksi_tools/console/settings_view.py
```diff
@@ -38,54 +33,19 @@
...
 
-SETTINGS_REVERT_TIMEOUT = 5
-SETTINGS_RETRY_TIMEOUT = 10
-BATCH_WINDOW = 10
+# TODO: maybe something more elegant..
+import sys
+sys.path.insert(0, 'lib/libsettings')
+# sys.path.insert(0, '/mnt/users/pasi/swiftnav/libsettings/python')
+
+from libsettings import Settings, settings_write_res_e   # noqa: E402
```
silverjam: Not sure I understand the `settings_write_res_e` naming, using lower case with underbars is for functions,  right?  Would suggest `SettingsWriteResponseCodes` or just `ResponseCodes`.

piksi_tools/console/settings_view.py
```diff
@@ -222,33 +182,43 @@ def revert_to_prior_value(self, name, old, new, error_value=SBP_WRITE_STATUS.TIM
...
-                    # regardless of which way setting is going, a restart is required
-                    else:
-                        self.settings.display_ins_output_hint()
+        if getattr(self, 'settings', None) is None:
+            return
+
+        if self.reverting:
+            return
+
+        threading.Thread(target=self._write_value, args=(old, new)).start()
```
silverjam: I'm assuming most of the logic in`TimedDelayStoppableThread` went into libsettings, otherwise we're losing a lot of what looks like retry logic.  I would appreciate some verbiage in the final commit message for this PR to mention this.

Also assuming that libsettings probably doesn't deal with threads, so we put it into a thread to prevent it from blocking the UI?  Do we want to make a task queue for settings writes instead or is it OK to spawn one thread per settings write?

piksi_tools/console/settings_view.py
```diff
@@ -505,41 +475,82 @@ def display_ins_output_hint(self):
...
-        # start a thread that will resend any read indexes that haven't come
-        self._restart_retry_thread()
+    def _settings_unconfirm_all(self):
+        for section in self.settings.keys():
+            for name in self.settings[section].keys():
+                self.settings[section][name].confirmed_set = False
+
+    def _settings_read_all(self):
+        self._settings_unconfirm_all()
+        threading.Thread(target=self._settings_read_all_thd).start()
```
silverjam: Let's make a task queue for these operations so we can be more disciplined about when threads are created.  If we need more than one thing to run at the same time we can use a thread pool to execute the tasks.

piksi_tools/console/settings_view.py
```diff
@@ -505,41 +475,82 @@ def display_ins_output_hint(self):
...
+
+    def _settings_read_all_thd(self):
+        slist = self.s_api.read_all()
```
silverjam: Suggest `slist` -> `settings` or `settings_list`.

piksi_tools/console/settings_view.py
```diff
@@ -505,41 +475,82 @@ def display_ins_output_hint(self):
...
+    def _settings_unconfirm_all(self):
+        for section in self.settings.keys():
+            for name in self.settings[section].keys():
+                self.settings[section][name].confirmed_set = False
+
+    def _settings_read_all(self):
+        self._settings_unconfirm_all()
+        threading.Thread(target=self._settings_read_all_thd).start()
+
+    def _settings_read_all_thd(self):
```
silverjam: Suggest `_settings_read_all_thd` -> `_read_all_thread`.

piksi_tools/console/settings_view.py
```diff
@@ -505,41 +475,82 @@ def display_ins_output_hint(self):
...
+        slist = self.s_api.read_all()
+
+        idx = 0
+
+        for s in slist:
```
silverjam: Suggest `s` -> `setting`.

piksi_tools/console/settings_view.py
```diff
@@ -862,39 +761,26 @@ def __init__(self,
                  gui_mode=True,
                  skip=False):
         super(SettingsView, self).__init__()
-        self.ordering_counter = 0
+        self.s_api = Settings(1234, link)
```
silverjam: Suggest `s_api` -> `settings_api` or `libsettings`, or `settings_instance`.

This used to default to SENDER_ID from `sbp.msg` from what I can tell (https://github.com/swift-nav/libsbp/blob/master/python/sbp/settings.py#L79).  Should we do the same here?

piksi_tools/console/settings_view.py
```diff
@@ -505,41 +475,82 @@ def display_ins_output_hint(self):
...
+                setting_type = None
+                setting_format = None
+            else:
+                setting_type, setting_format = fmt_type.split(':')
+
+            if section not in self.settings:
+                self.settings[section] = {}
+
+            # setting exists, we won't reinitilize it but rather update existing setting
+            dict_setting = self.settings[section].get(name, False)
```
silverjam: Suggest `dict_setting` -> `existing_setting`?

piksi_tools/console/settings_view.py
```diff
@@ -505,41 +475,82 @@ def display_ins_output_hint(self):
...
+            if dict_setting:
+                dict_setting.value = value
+                dict_setting.ordering = idx
+                if setting_type == 'enum':
+                    enum_values = setting_format.split(',')
+                    dict_setting.enum_values = enum_values
+                dict_setting.confirmed_set = True
+                continue
+
+            if setting_type is None:
```
silverjam: The `if` and `else` clause look the same, merge into `if setting_type != 'enum'`?

piksi_tools/console/settings_view.py
```diff
@@ -566,29 +577,53 @@ def _settings_export_to_file_button_fired(self):
...
-            try:
-                with open(file.path, "w") as f:
-                    parser.write(f)
-            except IOError as e:
-                print('Unable to export settings to file due to IOError: {}'.format(e))
-        else:  # No error message because user pressed cancel and didn't choose a file
-            pass
+
+        if file.open() != OK:
+            # No error message here because user likely pressed cancel when choosing file
```
silverjam: Is this always true?  Can we check if there was an error opening the file?  Maybe we should print a message here?

piksi_tools/console/settings_view.py
```diff
@@ -599,66 +634,76 @@ def _settings_import_from_file_button_fired(self):
...
-                    parser.read_file(f)
-            except configparser.ParsingError as e:  # file formatted incorrectly
-                print('Unable to parse ini file due to ParsingError: {}.'.format(e))
-                print('Unable to import settings to device.')
-                return
-            except IOError as e:  # IOError (likely a file permission issue)
-                print('Unable to read ini file due to IOError: {}'.format(e))
-                print('Unable to import settings to device.')
+        if file.open() != OK:
+            # No error message here because user likely pressed cancel when choosing file
```
silverjam: Same as above comment?

piksi_tools/console/settings_view.py
```diff
@@ -222,33 +182,43 @@ def revert_to_prior_value(self, name, old, new, error_value=SBP_WRITE_STATUS.TIM
...
-                        hint_thread = threading.Thread(
-                            target=self.settings._display_ins_settings_hint)
-                        hint_thread.start()
-                    # regardless of which way setting is going, a restart is required
-                    else:
-                        self.settings.display_ins_output_hint()
+        if getattr(self, 'settings', None) is None:
+            return
+
+        if self.reverting:
```
silverjam: I think it also makes sense to avoid firing a thread (or a task) if `old != new`.

piksi_tools/console/settings_view.py
```diff
@@ -599,66 +634,76 @@ def _settings_import_from_file_button_fired(self):
...
+                this_setting = this_section.get(setting, None)
+                if this_setting is None:
+                    self._import_failure_not_found(section, setting)
+                    return
+
+                if this_setting.value == value:
+                    continue
+
+                this_setting.write_event.clear()
+                this_setting.value = value
```
silverjam: I'm not convinced that using the GUI's method of updating the data is the best choice here.  If we just call the underlying `self._write_value` directly we can make this whole method blocking without the wait on `this_settings.write_event`.  When `self._value_changed` is fired we can avoid firing a thread if the value hasn't changed.

.appveyor.yml
```diff
@@ -12,6 +12,11 @@ environment:
       PYTHON_VERSION: 2.7
       PYTHON_ARCH: 32
 
+  APPVEYOR_RDP_PASSWORD: Libsettings911!
+
+init:
```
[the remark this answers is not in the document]
denniszollo: indeed, make sure to remove any remote desktop feature on appveyor after getting it working

.appveyor.yml
```diff
@@ -26,10 +31,22 @@ install:
...
   - cinst nsis
   - SET PATH=%PYTHON%;%PYTHON%\Scripts;%PATH%
+  - set PATH=C:\MinGW\bin;%PATH%
   - python --version
   # Upgrade to the latest pip.
   - '%CMD_IN_ENV% python -m pip install -U pip setuptools wheel tox'
   - powershell .\\tasks\\missing-headers.ps1
+  - md lib
+  - cd lib
+  - appveyor DownloadFile "https://raw.githubusercontent.com/swift-nav/libsettings/56ac4611e9fb148d7beaf469384247872f056f6d/python/libsettings-0.1.0.tar.gz"
```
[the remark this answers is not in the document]
denniszollo: Lets make sure that adding MinGW to the path doesn't somehow affect the pyinstaller / nsis builds.  We should ensure that the installer still works on windows 7 & 10 32 and 64 bit before merging.

.appveyor.yml
```diff
@@ -26,10 +31,22 @@ install:
...
+  - md libsettings
+  - tar -xzf libsettings-0.1.0.tar.gz -C libsettings --strip-components=1
+  - cd libsettings
+  # Hack to get rid of format specifier warnings emerging from the cythonized file
+  - C:\cygwin64\\bin\sed.exe -i 's/#define CYTHON_FORMAT_SSIZE_T "z"/#define CYTHON_FORMAT_SSIZE_T "I"/g' python/libsettings.c
+  - C:\cygwin64\\bin\sed.exe -i "s/zd /Id /g" python/libsettings.c
+  - python setup.py build_ext --inplace --compiler=mingw32
+  - cd ../..
```
denniszollo: What does this mean from a "pip install piksi_tools" perspective on a windows machine?  

Right now you can pip install the command line tools anywhere and they just work.  Custmoers use them to programmatically configure devices.  Does libsettings have its own command line executable that people should use instead?  

It seems to me that installing piksi_tools on windos is going  to get much harder with libsettings.  Is it possible?  If so, what is the procedure?  Customer's likely won't have mingw, cygwin, or sed; in that case, could they install  piksi_tools and get access to a settings command line client?

piksi_tools/console/settings_view.py
```diff
@@ -222,33 +182,43 @@ def revert_to_prior_value(self, name, old, new, error_value=SBP_WRITE_STATUS.TIM
...
         invalid_setting_prompt.run()
+        self.reverting = False
+
+    def _write_value(self, old, new):
+        if (old != new and old is not Undefined and new is not Undefined):
+            if type(self.value) == unicode:
+                self.value = self.value.encode('ascii', 'replace')
+            self.confirmed_set = False
+            res = self.settings.s_api.write(self.section, self.name, new)
+            if res == settings_write_res_e.SETTINGS_WR_OK:
```
denniszollo: What if the write response is never received?  In a crowded serial environment or over radio or a lossy link, it is often the case that you don't receive settings write response.  In that case, is the write retried?

piksi_tools/console/settings_view.py
denniszollo: Please make sure to test this thoroughly when a device is navigating with 5 Hz observations and default settings on a real serial port.  You cannot imagine the amount of testing I did on a real device when I added in the retry and logic for making sure a settings write took.  I think it is also important to test with firmware 2.0 that had a very lossy SBP interface due to nanomsg implementation.  I will plan to test on my end first thing next week to see if I notice anything odd.

.appveyor.yml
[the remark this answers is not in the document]
pmiettinen: This was added to get visibility to frozen builds. I will remove this before the merge.

.appveyor.yml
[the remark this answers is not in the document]
pmiettinen: @silverjam Tag.

@denniszollo OK.

.appveyor.yml
pmiettinen: No command line executable. I need to investigate what actually happens with `pip install piksi_tools`.

Few comments on the spot:

`cygwin` is only needed for `sed`. `sed` is needed only for getting rid of format specifier compatibility warnings from cython source error messages so not a hard requirement. I first tried to do the replacement with powershell tools but windows cmd line is very picky with the quotes and escape chars. I managed to get it working locally but not on appveyor environment. *enter sarcastic comments about unix shell vs win cmd*

`mingw` is a must if we want to have `c99` in libsettings, for a moment I already started writing `c89`/`c90` compatible code but... it just feels so dumb. In case of Python3 windows might be using newer compiler that actually supports `c99`, not sure about this.

piksi_tools/console/settings.yaml
pmiettinen: This wasn't about `libsettings`. It was about my environment. Anyway there was also another reason to do this change; to unify the note with the next setting note that has the same semantic.

.appveyor.yml
pmiettinen: I did some investigations. It is unclear for me if `python -m piksi_tools.console.console` is expected to work on current `master` after running `pip install piksi_tools` on Windows. See slack conversation I started https://snav.slack.com/archives/C0F4WHXK5/p1546249486001200

If use case is not expected to work then this issue is also irrelevant regarding `settings_view.py`. Although this means that settings.py tool can't be upgraded to use libsettings at this stage.

pmiettinen: I think the closest thing what you mean by libsettings command line client is `settings_client.py` in https://github.com/swift-nav/libsettings/pull/3

Nonetheless it requires building `libsettings` first. Adding win build support is possible.

I suggest we leave `piksi_tools/settings.py` tool as is for now and in parallel work to get similar functionality and ease of use with `libsettings/settings_client.py` in subsequent PRs.

piksi_tools/console/settings_view.py
pmiettinen: More elaborate; write is done only once and libsettings blocks waiting for the answer for 0.5 seconds for 5 times, so 2.5 seconds in total, this behavior was inherited from legacy code. If response is not received `SETTINGS_WR_TIMEOUT` is returned.

piksi_tools/console/settings_view.py
pmiettinen: Rather `old == new` you mean?

piksi_tools/console/settings_view.py
pmiettinen: I might still be on the vacation mode as I don't see the retry logic for writing? Only in the indexed read case which admittedly needs to be added to libsettings.

Misc comments:
- libsettings doesn't support multithreading currently
- Thread is created to avoid blocking UI

piksi_tools/console/settings_view.py
pmiettinen: Could you elaborate what exactly you mean by task queue?

piksi_tools/console/settings_view.py
pmiettinen: Correction: libsettings waits for the response for 0.5 seconds for 5 times, so 2.5 seconds in total; write request is sent again on each try; this behavior was inherited from legacy code.

If response is not received `SETTINGS_WR_TIMEOUT` is returned.

piksi_tools/console/settings_view.py
pmiettinen: Write retry logic is in libsettings, see comment above: https://github.com/swift-nav/piksi_tools/pull/938/files#r244578599

piksi_tools/console/settings_view.py
pmiettinen: Agreed.

piksi_tools/console/settings_view.py
pmiettinen: Fine by me if no downsides emerge.

piksi_tools/console/settings_view.py
pmiettinen: Done.

piksi_tools/console/settings_view.py
pmiettinen: Done.

piksi_tools/console/settings_view.py
pmiettinen: Done.

piksi_tools/console/settings_view.py
pmiettinen: Done.

piksi_tools/console/settings_view.py
pmiettinen: Renamed to `settings_api`. Need to verify the ID issue.

piksi_tools/console/settings_view.py
pmiettinen: Renamed to `SettingsWriteResponseCodes`

silverjam: 👍

.appveyor.yml
silverjam: I think that is fine, but we'd need to eventually get libsettings able to install via pip for any"pip install" based deployment to work, right?  This would mean deploying to PyPI.

piksi_tools/console/settings.yaml
silverjam: 👍

piksi_tools/console/settings_view.py
silverjam: A thread (or pool of threads) that waits on a shared queue and executes tasks (functions) as they are enqueued.  Which (coincidentally?) looks like exactly what you've already implemented! 😄

piksi_tools/console/settings_view.py
silverjam: https://swift-nav.atlassian.net/browse/ESD-987

piksi_tools/console/settings_view.py
silverjam: Yes

.appveyor.yml
silverjam: I think this captures the issue? https://swift-nav.atlassian.net/browse/ESD-988

pmiettinen: Yes and yes (Y)

pmiettinen:
I have cleared the commit history and done some basic sanity testing. One defect has been found and described in the original pull request message.

@denniszollo - I guess you are the ultimate gatekeeper for this. As said under non-stressed conditions everything seems to be in order except for the one issue mentioned. Did you have time to test this with lossy SBP link or do you have time to do so with this latest version?

requirements.txt
```diff
@@ -1,3 +1,4 @@
+--extra-index-url https://testpypi.python.org/pypi
```
pmiettinen: Remove from the final version.

setup.py
```diff
@@ -121,7 +121,7 @@ def scmtools_parse(root,
 
     with open(cwd + '/requirements.txt') as fp:
 
-        INSTALL_REQUIRES = [L.strip() for L in fp if not L.startswith('git+')]
+        INSTALL_REQUIRES = [L.strip() for L in fp if (not L.startswith('git+') and not L.startswith('--extra-index-url'))]
```
tn-exafore: Is this a development-time hack that is not needed anymore?

pmiettinen: It has a niche case if you want to install from `test.pypi.org`.

piksi_tools/console/settings_view.py
```diff
@@ -14,15 +14,12 @@
 import threading
 import time
 import configparser
-from enum import IntEnum, unique
+
+from queue import Queue
```
tn-exafore: This is one of the modules renamed in python 2.7->3. Did you test both versions? For safety, `from six.moves.queue import Queue`

pmiettinen: Yes it passes tests for both in CI. But I'll make the change.

pmiettinen: IIRC `Queue` is Python2 only. `queue` works for both.

piksi_tools/console/settings_view.py
```diff
@@ -507,41 +492,76 @@ def display_ins_output_hint(self):
...
+        self._settings_unconfirm_all()
+        self.workqueue.put(self._read_all_thread)
+
+    def _read_all_thread(self):
+        settings_list = self.settings_api.read_all()
+
+        idx = 0
+
+        for setting in settings_list:
+            section = setting['section']
```
tn-exafore: I presume that the libsettings side returns strings that work in both python 2 and 3... have you tested it with both 2.7 and 3.5?

pmiettinen: Doesn't seem to work under Python 3.5.0

```
pasi@pasi-linux:~/swiftnav/libsettings/python$ python settings_client.py --tcp -p 192.168.1.25:55555
('192.168.1.25', 55555)
Creating Settings
b'Building settings framework'
  File "settings_client.py", line 105, in <module>
    main()
  File "settings_client.py", line 90, in main
    print("solution.elevation_mask =", s.read("solution", "elevation_mask"))
  File "python/libsettings.pyx", line 160, in libsettings.Settings.read
    bytes(section),
TypeError: string argument without an encoding
pasi@pasi-linux:~/swiftnav/libsettings/python$ python --version
Python 3.5.0
```

pmiettinen: https://github.com/swift-nav/libsettings/pull/31

piksi_tools/console/console.py
```diff
@@ -715,11 +710,16 @@ def update_serial():
                 if self.networking_view.connect_when_uuid_received:
                     self.networking_view._connect_rover_fired()
 
+            skip_settings_read = False
+            if 'mode' in self.connection_info:
+                if self.connection_info['mode'] == 'file':
+                    skip_settings_read = True
+
```
denniszollo: those errors when testing from a sbp file always bothered me; thanks for fixing.

piksi_tools/console/settings_view.py
```diff
@@ -223,34 +199,39 @@ def revert_to_prior_value(self, name, old, new, error_value=SBP_WRITE_STATUS.TIM
                  "   Error Value: {2}")
         invalid_setting_prompt.text = invalid_setting_prompt.text.format(self.name, new, error_value)
         invalid_setting_prompt.run()
+        self.reverting = False
```
denniszollo: Are we sure we want to keep reverting true for this long time?  I feel like there were corner cases on the pop-up that caused odd behavior and that is why I turned reverting back off as soon as the old value was written to setting.  If you've adequately tested, I'm okay with the change, but I would prefer to unset .reverting as soon as we have done the revert to prior value.

piksi_tools/console/settings_view.py
```diff
@@ -223,34 +199,39 @@ def revert_to_prior_value(self, name, old, new, error_value=SBP_WRITE_STATUS.TIM
...
+
+    def _write_value(self, old, new):
+        if (old != new and old is not Undefined and new is not Undefined):
+            self.confirmed_set = False
+            res = self.settings.settings_api.write(self.section, self.name, new)
```
denniszollo: Does this just block while the api waits for a response?  What does that do to the UI?  Have you tested over a lossy connection, or a slow connection?

denniszollo:
With respect to installing the libsettings dependency, I see that "setup.py develop" works, but I tried "pip install -e ./", which is my usual way to setup console, and that failed. Is this by design, so that users who only want the command line utilities can get them through pip without installing libsettings?

denniszollo:
Should we add "libsettings" to requirements_gui.txt ?

piksi_tools/console/settings_view.py
denniszollo: I have done some quick testing now on MacOs and see that it seems responsive when changing a setting even over a lossy connection, so I am satisfied.

piksi_tools/console/settings_view.py
```diff
@@ -343,11 +324,12 @@ class SettingsView(HasTraits):
...
     gui_mode : bool
       ??? (defaults to True)
-    skip : bool
+    skip_read : bool
       Skip reading of the settings (defaults to False). Intended for
-      use when reading from network connections.
+      use when reading from network connections or file.
     """
     show_auto_survey = Bool(False)
+    enable_buttons = Bool(True)
```
denniszollo: The current implementation prevents someone from resetting to default settings in the case that the settings could not be discovered.  There  are a lot of cases where settings might not be able to be discovered and we shouldn't neuter any functionality if that happens.

I don't think the "refresh" button or the "reset to default settings" button should be turned off when discovering settings.  I for one use both often when the settings cannot be discovered.

piksi_tools/console/settings_view.py
```diff
@@ -433,7 +417,7 @@ def _selected_setting_changed(self):
 
     def _expert_changed(self, info):
         try:
-            self.settings_display_setup(do_read_finished=False)
+            self.settings_display_setup(finish=False)
```
denniszollo: finish" is less clear to me.  What does it mean? "do_read_finished" behaves as named.

piksi_tools/console/settings_view.py
```diff
@@ -507,41 +491,76 @@ def display_ins_output_hint(self):
...
+    def _settings_unconfirm_all(self):
+        # Clear the tabular editor
+        del self.settings_list[:]
+        for section in self.settings.keys():
+            for name in self.settings[section].keys():
+                self.settings[section][name].confirmed_set = False
+
+    def _settings_read_all(self):
+        self._settings_unconfirm_all()
+        self.workqueue.put(self._read_all_thread)
```
denniszollo: this is much slower than the old implementation,  It almost always takes 10 - 20 seconds to discover all the settings.  Why is this?  Is it necessary that this is slow?

I also tried discovering settings over a lossy connection and I got some really strange stdout output:
image and the settings were never discovered.

piksi_tools/console/settings_view.py
```diff
@@ -601,66 +644,72 @@ def _settings_import_from_file_button_fired(self):
...
+            for setting, value in settings.items():
+                this_setting = this_section.get(setting, None)
+                if this_setting is None:
+                    self._import_failure_not_found(section, setting)
+                    return
+
+                if this_setting.value == value:
+                    continue
+
+                this_setting._write_value(this_setting.value, value)
```
denniszollo: What happens if user imports a setting that breaks communication? How long will it take to timeout?

piksi_tools/console/settings_view.py
```diff
@@ -601,66 +644,72 @@ def _settings_import_from_file_button_fired(self):
...
-        else:
-            pass  # No error message here because user likely pressed cancel when choosing file
+            for setting, value in settings.items():
+                this_setting = this_section.get(setting, None)
+                if this_setting is None:
+                    self._import_failure_not_found(section, setting)
+                    return
+
+                if this_setting.value == value:
+                    continue
```
denniszollo: Not writing a value here relies heavily on the console's idea of what a current setting's value is, which has no basis in reality and is basically just a cache.  If we are going to ignore lines based upon the console's cache, we should require a settings.read_all() immediately before importing a file or have some other way to ensure that the cache is updated for the currently connected device before choosing ignore a settings write.  If I ask to write a bunch of settings from a file, I kind of expect them to be written and I don't want the console to outsmart me.

A use case that I have  instructed customers to use to configure devices in bulk is to use the same console session to configure many devices by unplugging and replugging serial into each device one by one.  In that use case, this will be broken because the console will ignore the file based upon its cache of settings values.  Similarly, If a device has restarted, but the console did not receive the STARTUP_MSG triggering a refresh of the cache, the console could ignore a settings write based upon the settings value at the last power-up.

What is the harm in writing a setting even if the value is the same?

denniszollo: This also kind of masks the fact that reading settings from a file has taken a pretty big performance hit compared with the old version.  The reason I did not re-use the GUI logic for changing a setting when reading from a file was so you could import an entire file in seconds and verify all the settings in parallel.  I went through and changed the value of every single setting on a Piksi Multi and see that it takes 30 seconds or more in some cases to import that file.  When it takes a long time to import the file, the UI experience isn't very good.  As a user, you are completely unaware that anything is happening until the pop-up appears. 

For a reproduction scenario, over serial at 115200 baud, change all of the settings in the NMEA group to "0", then import a config.ini file with the NMEA settings group set to the default value.  These 8 settings alone take 10 seconds on my mac before the success pop-up.  Imagine a larger file import, the time could get quite long indeed.

If we are going to incur this big performance hit, I almost want the GUI to block until the settings import is complete. Is there some way to make things faster to get closer to the old implementation in performance?

piksi_tools/console/settings_view.py
```diff
@@ -712,203 +742,56 @@ def settings_display_setup(self, do_read_finished=True):
...
+        if finish:
+            self.finish_read()
 
     def piksi_startup_callback(self, sbp_msg, **metadata):
-        self._settings_read_by_index()
-
-    def set(self, section, name, value):
-        self.link(
-            MsgSettingsWrite(setting=b'%s\0%s\0%s\0' % (section, name, value)))
+        self._settings_read_all()
```
denniszollo: Perhaps we can call a read_all if we are getting a never before seen SBP Sender_ID as well to avoid the console's local cache not matching the device that is plugged in.

piksi_tools/console/settings_view.py
pmiettinen: I have tested changing the unsupported `qzss_acquisition_enabled` setting and have not encountered any anomalies this far. The previous version did reverting in a new thread which can complicate things. In here all happens in the same work thread if the write fails.

piksi_tools/console/settings_view.py
pmiettinen: All reads and writes happen in separate `WorkQueue` threads leaving the GUI free to do its job.

piksi_tools/console/settings_view.py
pmiettinen: I am not sure if I follow. Buttons are disabled only when there's read or write already happening preventing the user for example from spamming the refresh button. Reads and writes will all timeout eventually to make the buttons enabled again.

piksi_tools/console/settings_view.py
pmiettinen: With current defines each request is tried 5 times each having 500ms timeout so 2.5 seconds total.

piksi_tools/console/settings_view.py
pmiettinen: I am not 100% familiar with the earlier solution but seems like it does burst-like read request batches each batch containing 10 requests. So in away there's 10 parallel requests compared to current `libsettings` architecture allowing only one request in the air per client at a time. This is probably amplifying the latency effect of the req-resp roundtrip over serial line.

Furthermore the original solution keeps sending the indexed read requests until all indexes are covered which in the worst case can take forever.

How valid is that use case where the buffers are full?

The mystic message output is the contents of the request, it can be removed.

pmiettinen: Looking at the code I am not convinced the earlier solution works as expected especially during the last batch. `settings_read_by_index_done_callback` stops the retry thread without checking if there's still pending read requests unanswered.

pmiettinen: Here's a proof of concept level [PR for multithreading support in `libsettings`](https://github.com/swift-nav/libsettings/pull/33). It can support 1..n worker threads for doing indexed read for settings.

I didn't have time to test over serial line yet, but over network the refresh of settings was pretty much instant. Of course the lag wasn't so noticeable over network to start with.

Anyway, this code is still at least few days from production level quality and some design decisions still needs considering. @benjaminaltieri what do you think is the priority of this work compared to the `libsbp` Numba integration?

piksi_tools/console/settings_view.py
pmiettinen: Agreed that the GUI thread is not good place to do the writes. Actually I was surprised that this didn't go through the `WorkQueue`.

piksi_tools/console/settings_view.py
benjaminaltieri: Let's merge what you have here. We can follow on with the multithreading when appropriate. Thanks for putting that together.

piksi_tools/console/settings_view.py
denniszollo: I think that the buttons are disabled while the initial settings discovery is taking place.  Is that not the case?

denniszollo:
console_settings

denniszollo:
[console_settings.mov.zip](https://github.com/swift-nav/piksi_tools/files/3014767/console_settings.mov.zip)

benjaminaltieri:
I've now run locally and have observed a 1-2 second delay on initial settings read by index when connecting to the piksi, which is not huge but is perceptible enough that we should hold until there is a solution

piksi_tools/console/settings_view.py
pmiettinen: I see, so the variable shall be initialized to `False`.

pmiettinen: Previous comment was incorrect. Buttons are enabled at start, but disabled during the `read_all` and it takes 2.5s * amount_of_settings to timeout if there's no connection. It is questionable what's the use of enabled buttons if there's no connection.

pmiettinen: But I'll revert the button disable/enable feature since it's not the point of this PR.

piksi_tools/console/settings_view.py
pmiettinen: reverted

piksi_tools/console/settings_view.py
pmiettinen: There's `old != new` guard also in the original `_value_changed()` so I don't think the functionality has changed in that regard.

pmiettinen: I'll remove all the `old` vs `new` guard.

piksi_tools/console/settings_view.py
pmiettinen: I presume this would be a new feature? If so I propose to do it in a separate PR.

pmiettinen:
> With respect to installing the libsettings dependency, I see that "setup.py develop" works, but I tried "pip install -e ./", which is my usual way to setup console, and that failed. Is this by design, so that users who only want the command line utilities can get them through pip without installing libsettings?

This is because currently `libsettings` requirement in the `requirements.txt` is from github dev branch and `setup.py` filters those out.

`INSTALL_REQUIRES = [L.strip() for L in fp if (not L.startswith('git+') and not L.startswith('--extra-index-url'))]`

piksi_tools/console/settings_view.py
```diff
@@ -568,29 +578,101 @@ def _settings_export_to_file_button_fired(self):
...
+        ret = self.settings_api.write_all(settings_to_write)
+
+        success = True
+
+        for (error, section, name, value) in ret:
+            if error and error != SettingsWriteResponseCodes.SETTINGS_WR_READ_ONLY:
+                self._import_failure_write(error, section, name)
+                success = False
+            else:
+                self.settings.get(section, None).get(name, None).value = value
```
pmiettinen: I think this triggers redundant write through `_value_changed` -> filter it out.

piksi_tools/console/settings_view.py
denniszollo: yes this was just me brainstorming; no action required for this PR.

piksi_tools/console/settings_view.py
```diff
@@ -568,29 +575,103 @@ def _settings_export_to_file_button_fired(self):
...
+            "  Settings import from file complete.  Click OK to save the settings  \n" \
+            "  to the device's persistent storage.  \n"
+        confirm_prompt.run(block=False)
+
+    def _import_fail(self):
+        confirm_prompt = prompt.CallbackPrompt(
+            title="Failed to import settings from file",
+            actions=[prompt.close_button])
+        confirm_prompt.text = "\n" \
+            "  Verify that config file is not overwriting active connection settings.  \n"
```
pmiettinen: Now that file importing is truly sending every setting from the file, I found it necessary to add this warning. @denniszollo

pmiettinen:
Regarding getting `pip install -e ./` to work:
- tag libsettings-v0.1.11
- push it to PyPI after we are convinced it's stable/working
- update `requirements.txt`

pmiettinen:
In case of device/connection failure during read/write all, there'll be unnecessary warning spam in the logging window. TODO: suppress these.

piksi_tools/console/settings_view.py
denniszollo: That is a good warning.

PR #941, opened 2018-12-01
[ESD-906] Update Settings for 2.2 release

anth-cole:


nsirola:
Btw, the output modes' help textas do not yet list `RTCMv3 OUT` anywhere.

piksi_tools/console/settings.yaml
```diff
@@ -2443,3 +2453,98 @@
...
+    "NMEA OUT" has no effect for CAN
+
+    "RTCMv3 IN" has no effect for CAN
+
+- group: can0
+  name: enabled_sbp_messages
+  expert: true
+  type: string
+  units: N/A
+  default value: '23,29,65,72,74,81,117,134,136,144,163,165, 166,167,171,175,181,185,187,188,189,190, 257,258,259,520,522,524,526,527,528,1025, 2304,2305,2306,30583,65280,65282,65535'
```
pmiettinen: This is probably not what we want. @switanis to comment.

piksi_tools/console/settings.yaml
```diff
@@ -2443,3 +2453,98 @@
...
+  default value: 'true'
+  readonly: false
+  Description: Enable metric logging to file
+  Notes:
+
+- group: system
+  name: resource_monitor_update_interval
+  expert: true
+  type: integer
+  units: Hz
```
silverjam: Unit is seconds, not hz

piksi_tools/console/settings.yaml
```diff
@@ -2443,3 +2453,98 @@
...
+  type: enum
+  units: N/A
+  default value: 'Disabled'
+  readonly: false
+  enumerated possible values: SBP,NMEA OUT,RTCMv3 IN,RTCMv3 Out
+  Description: Communication protocol for CAN client 0. The client will send packets to a CAN bus.
+  Notes: |
+    "SBP" configures the interface to transmit messages specified in the 'enabled_sbp_messages' setting.
+
+    "NMEA OUT" has no effect for CAN
```
silverjam: This isn't technically true as far as I know, right @pmiettinen?  You can configure the CAN ports to emit any protocol that we support.

piksi_tools/console/settings.yaml
silverjam: I think we only want 72, 74, 117, 522, 527 -- @switanis anything else?

piksi_tools/console/settings.yaml
```diff
@@ -2443,3 +2453,98 @@
   default value: N/A
   readonly: true
   Description: inertial navigation system build date
+
+- group: metrics_daemon
+  name: metrics_update_interval
+  expert: true
+  type: integer
+  units: Hz
```
silverjam: Unit is seconds, not hz

PR #946, opened 2018-12-04
added checks for reading from settings dictionary in update_view

jkretzmer:
previous build failed to upgrade non-ins hardware due to 'ins' setting group not being populated. changed the settings read to check for the presence of the 'ins' group and 'output_mode' setting before checking the value.


denniszollo:
lets plan to port to v2.2 only if unpleasant upgrade experience is reproduced.

PR #947, opened 2018-12-04
remove WRITE_RESP message from udp/tcp_client [ESD-922] (#944)

yizow:
A temporary solution to ESD-922. Changing a setting on one piksi would
also send a WRITE_RESP to a second piksi, also causing an unintentional
setting change there.


yizow:
This change is specifically to disallow making settings write over the tcp_client interface. 
A piksi should never be writing settings on another piksi.

PR #954, opened 2019-01-07
Add tentative python 3 support to the most important command line tools [EX-477]

tn-exafore:
This adds tentative python 3 support to the most important command line apps:
* fileio.py
* bootload_v3.py
* settings.py
* serial_link.py
while also maintaining their python 2 support. Other command line apps are not updated to python 3 for now and are semi-abandoned (and in fact there could be also other issues in them already as they don't seem to be well maintained).

The python 3 support is preliminary and can in principle be tested on top of the WIP libsbp python 3 branch https://github.com/swift-nav/libsbp/commits/silverjam/py3 with two notes:
* needs rebasing (I tested using a locally rebased copy but didn't commit the changes yet, to be done in EX-476)
* serial_link.py's log-to-file functionality needs a minor fix to a logger class in libsbp (likewise an uncommitted local change at the moment)

The python 2 support works on top of the normal libsbp master branch.

Minor changes to the cli tools may be necessary after or in conjunction with EX-476 (finalizing libsbp python 3 support).

### DIFF
--- a/piksi_tools/fileio.py
+++ b/piksi_tools/fileio.py
@@ -10,6 +10,7 @@
 # WARRANTIES OF MERCHANTABILITY AND/OR FITNESS FOR A PARTICULAR PURPOSE.
 
 from __future__ import absolute_import, print_function
+from future.builtins import bytes  # makes python 2 bytes more similar to python 3
 
 import copy
 import itertools
@@ -229,12 +230,12 @@ class FileIO(object):
 
         Parameters
         ----------
-        filename : str
+        filename : bytes
             Name of the file to read.
 
         Returns
         -------
-        out : str
+        out : bytearray
             Contents of the file.
         """
         offset = 0
@@ -265,18 +266,18 @@ class FileIO(object):
             sorted_buffers = sorted(closure['buf'].items(), key=lambda kv: kv[0])
             return bytearray(itertools.chain.from_iterable(kv[1] for kv in sorted_buffers))
 
-    def readdir(self, dirname='.'):
+    def readdir(self, dirname=b'.'):
         """
         List the files in a directory.
 
         Parameters
         ----------
-        dirname : str (optional)
+        dirname : bytes (optional)
             Name of the directory to list. Defaults to the root directory.
 
         Returns
         -------
-        out : [str]
+        out : [bytes]
             List of file names.
         """
         files = []
@@ -292,10 +293,11 @@ class FileIO(object):
             reply = MsgFileioReadDirResp(reply)
             if reply.sequence != seq:
                 raise Exception("Reply FILEIO_READ_DIR doesn't match request (%d vs %d)" % (reply.sequence, seq))
-            chunk = str(bytearray(reply.contents)).rstrip('\0')
+            chunk = bytes(reply.contents).rstrip(b'\0')
+
             if len(chunk) == 0:
                 return files
-            files += chunk.split('\0')
+            files += chunk.split(b'\0')
 
     def remove(self, filename):
         """
@@ -303,7 +305,7 @@ class FileIO(object):
 
         Parameters
         ----------
-        filename : str
+        filename : bytes
             Name of the file to delete.
         """
         msg = MsgFileioRemove(filename=filename)
@@ -315,9 +317,9 @@ class FileIO(object):
 
         Parameters
         ----------
-        filename : str
+        filename : bytes
             Name of the file to write to.
-        data : str
+        data : bytearray
             Data to write
         offset : int (optional)
             Offset into the file at which to start writing in bytes.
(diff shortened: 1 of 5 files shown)


piksi_tools/fileio.py
```diff
@@ -470,21 +479,24 @@ def main():
...
+                    f.write(raw_filename(args.write[1]), bytearray(open(args.write[0], 'rb').read()))
                 elif args.read:
                     if len(args.read) not in [1, 2]:
                         sys.stderr.write("Error: fileio read requires either 1 or 2 arguments, SOURCE and optionally DEST.")
                         sys.exit(1)
-                    data = f.read(args.read[0])
+                    data = f.read(raw_filename(args.read[0]))
                     if len(args.read) == 2:
-                        with open(args.read[1], ('wb' if args.hex else 'w')) as fd:
+                        with open(args.read[1], ('w' if args.hex else 'wb')) as fd:
```
pmiettinen: Why the switch?

piksi_tools/fileio.py
```diff
@@ -470,21 +479,24 @@ def main():
...
                             fd.write(hexdump(data) if args.hex else data)
                     else:
-                        print(hexdump(data) if args.hex else data)
+                        if args.hex:
```
pmiettinen: combine to `elif`

piksi_tools/settings.py
```diff
@@ -239,16 +245,20 @@ def write_from_file(self, output, verbose=False):
...
-        self.read_response_wait_dict[(section, setting)] = value
+            b'\0')[:4]
+        self.read_response_wait_dict[(
+            section.decode(KEY_ENCODING), setting.decode(KEY_ENCODING))] = value.decode(VALUE_ENCODING)
 
     def _settings_list_callback(self, sbp_msg, **metadata):
-        section, setting, value, format_type = sbp_msg.payload[2:].split(
-            '\0')[:4]
+        section_b, setting_b, value_b, format_type_b = sbp_msg.payload[2:].split(
+            b'\0')[:4]
```
pmiettinen: keep the statement on line line if possible

piksi_tools/fileio.py
tn-exafore: Because I think the original order was actually incorrect: if we want to read a binary file from the device and write a copy locally as a hexdump, the output is a text file but if we want to write the local copy in original binary format, it's binary. This didn't actually matter in python2 because a text string could contain binary data that could be written into a file open in text mode but in python3 it matters.

piksi_tools/fileio.py
tn-exafore: neatified in a fix commit

piksi_tools/settings.py
tn-exafore: the other occurrence becomes a slightly over-long line but there are also others anyway in the same file so probably it's neater your way. Done in the review fix commit.

piksi_tools/fileio.py
```diff
@@ -393,11 +395,11 @@ def print_dir_listing(files):
 
     Parameters
     ----------
-    files : [str]
+    files : [bytes]
         List of file names in the directory.
     """
     for f in files:
-        print(f)
+        print(f.decode('ascii', 'replace'))
```
silverjam: I think we can safely make this `'utf8'` instead of `'ascii'` to add a little bit of future proofing.

```suggestion
        print(f.decode('utf8', 'replace'))
```

piksi_tools/fileio.py
```diff
@@ -455,6 +457,13 @@ def get_args():
     return parser.parse_args()
 
 
+def raw_filename(str_filename):
+    """Return a filename in raw bytes from a command line option string."""
+    # Non-unicode characters/bytes in the command line options are decoded by
+    # using 'surrogateescape' and file system encoding, and this reverts that.
```
silverjam: Is there a reference for this?  Can we link it?

piksi_tools/fileio.py
silverjam: 👍

piksi_tools/fileio.py
```diff
@@ -471,21 +480,23 @@ def main():
...
+                    data = f.read(raw_filename(args.read[0]))
                     if len(args.read) == 2:
-                        with open(args.read[1], ('wb' if args.hex else 'w')) as fd:
+                        with open(args.read[1], ('w' if args.hex else 'wb')) as fd:
                             fd.write(hexdump(data) if args.hex else data)
+                    elif args.hex:
+                        print(hexdump(data))
                     else:
-                        print(hexdump(data) if args.hex else data)
+                        print(data.decode('ascii', 'replace'))
```
silverjam:
```suggestion
                        print(data.decode('utf8', 'replace'))
```

piksi_tools/fileio.py
tn-exafore: I added two references:
* https://www.python.org/dev/peps/pep-0383/
* https://docs.python.org/3/library/os.html#file-names-command-line-arguments-and-environment-variables

piksi_tools/fileio.py
tn-exafore: Pondered a bit more on this non-utf-8 caller terminal situation and came up with a slightly complex solution, added as a separate commit: 
1. first decode device utf-8 bytes into python's internal string format
2. then encode that string to stdout's encoding (bytes) with error replacement
3. then decode back into python's string format. This string can be printed out with the print command

This has the advantage that no gibberish is printed out even if the user's system does not support utf-8 as long as locale envs are configured properly. If unicode characters from the device can be represented in the user's encoding, they will be, otherwise they will be replaced by question marks.

PR #955, opened 2019-01-11
Documentation for decimating raw IMU messages by factor of 50 [ESD-1014]

cbeighley:
Companion PR to https://github.com/swift-nav/piksi_buildroot/pull/1113

If you don't merge this, note that the default whitelist settings documentation in settings.yaml needs to get updated (done in this PR).

### DIFF
--- a/piksi_tools/console/settings.yaml
+++ b/piksi_tools/console/settings.yaml
@@ -824,7 +824,7 @@
   expert: false
   type: string
   units: N/A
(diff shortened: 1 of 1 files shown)


denniszollo:
@cbeighley Did we change this for 2.3 such that we need to update this setting?

silverjam:
@denniszollo we didn't enable the IMU by default in 2.3 afaik, @benjaminaltieri ?

benjaminaltieri:
@denniszollo @silverjam we implemented a variant that only activates IMU by default if INS is active (PR describing that here: https://github.com/swift-nav/piksi_buildroot/pull/1202)

silverjam:
Close?

PR #958, opened 2019-01-14
Clean up command line handling of cli tools a bit [EX-477]

tn-exafore:
* other cli tools "inherited" logging-related and --reset
  command line options from serial_link.py but didn't
  actually use them. Remove the unused ones by making them optional
* remove the unneeded append and decorate log args
* also add the -t for --tcp shorthand to fileio.py similarly
  to the other tools
* fix the timeout parameter in settings.py


piksi_tools/settings.py
```diff
@@ -317,8 +317,9 @@ def get_args():
                      "  2: Improper usage")
     parser.add_argument(
         "--timeout",
+        type=float,
         default=DEFAULT_TIMEOUT_SECS,
-        help="specify the timeout for settings reads.")
+        help="timeout for settings calls in s (default: %.1f)." % DEFAULT_TIMEOUT_SECS)
```
pmiettinen: `in seconds` or `timeout for settings calls [s]`

piksi_tools/settings.py
```diff
@@ -330,7 +331,7 @@ def get_args():
         'save', help='save all the current settings to flash.')
 
     reset = subparsers.add_parser(
-        'reset', help='reset the device after the action.')
+        'reset', help='reset the device to default settings.')
```
pmiettinen: default settings?

pmiettinen: I think this can be misinterpreted as factory defaults which I think is not the case?

piksi_tools/action_logger.py
```diff
@@ -264,16 +262,11 @@ def main():
         with Handler(Framer(driver.read, driver.write, args.verbose)) as link:
             # Logger with context
             with sl.get_logger(args.log, log_filename) as logger:
-                # Append logger iwth context
-                with sl.get_append_logger(append_log_filename,
-                                          tags) as append_logger:
                     # print out SBP_MSG_PRINT_DEP messages
                     link.add_callback(sl.printer, SBP_MSG_PRINT_DEP)
```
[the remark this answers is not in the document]
tn-exafore: yeah, I was lazy. Fixed.

piksi_tools/settings.py
tn-exafore: clarified

piksi_tools/settings.py
tn-exafore: it's indeed a reset to factory defaults AFAIK. Clarified.

piksi_tools/serial_link.py
```diff
@@ -73,68 +74,63 @@ def base_cl_options(override_arg_parse=None, add_help=True):
...
-        "--verbose",
+        "-t",
+        "--tcp",
         action="store_true",
-        help="print extra debugging information.")
+        default=False,
+        help="Use a TCP connection instead of a local serial port. \
+                      If TCP is selected, the port is interpreted as host:port"
+    )
     parser.add_argument(
```
denniszollo: I think we can safely remove the --ftdi option.  It isn't in use by anyone to my knowledge.  Doesn't need to be done as part of this effort but I thought I'd mention it.

piksi_tools/serial_link.py
```diff
@@ -381,21 +357,19 @@ def main(args):
...
-                        with Handler(
-                                Framer(http.read, http.write,
-                                       args.verbose)) as slink:
-                            Forwarder(slink, swriter(link)).start()
-                            run(args, link)
-                else:
-                    run(args, link)
+            link.add_callback(printer, SBP_MSG_PRINT_DEP)
+            link.add_callback(log_printer, SBP_MSG_LOG)
+            Forwarder(link, logger).start()
```
denniszollo: I think we can also remove this HTTP Forwarder feature longterm as well.  Doesn't need to be done as part of this effort but it isn't in use to my knowledge and causes us to import requests just to run serial_link.py even if it isn't in use.

denniszollo: This would allow us to remove a few more command line options to serial link and harmonize serial link's initialization with the other tools

tn-exafore: good pointers. I created a follow-up task to do these at some point in the future, after the python 3 update: https://swift-nav.atlassian.net/browse/ESD-1035

PR #959, opened 2019-01-14
Streamline settings cli tool write handling [EX-477]

tn-exafore:
This PR is an enhancement suggestion and depending on review comments, it can be merged, merged after minor or major changes or even rejected. A couple of points worth considering:
* I think (based on code and some discussion with Dennis) that the existing write implementation in settings.py is essentially a hack from a time when the write response msg didn't exist. It has been slightly modified to take into account write responses but not thoroughly. This PR makes "a more thorough" attempt.
* this PR drops retry attempts on writes. As I fixed the --timeout command line option recently, I think that's a better way to handle possibly failing writes due to network delays. Did the retries have any other value?
* there's no "NTRIP must be disabled to modify settings" warning anymore if attempting to write ntrip settings when it is enabled as that came from the log messages (a hack that resulted also in lots of unrelated junk being printed out). Instead, there's a bit more generic read-only warning that comes from the "proper" channel, i.e., write response
* the old implementation did some heuristic value checks in the confirm-by-reread method that the proposed one doesn't. I don't think those are needed on the client side: any failure checks should be done on the device firmware side and return write response with status SETTINGS_WR_VALUE_REJECTED or SETTINGS_WR_PARSE_FAILED, which this PR does check for
* thanks to streamlining, writing a big bunch of settings by write_from_file subcommand is roughly 6-10 times faster than in the old implementation

### DIFF
--- a/piksi_tools/settings.py
+++ b/piksi_tools/settings.py
@@ -322,7 +322,7 @@ def main():
     command = args.command
     return_code = 0
     driver = serial_link.get_base_args_driver(args)
-    with Handler(Framer(driver.read, driver.write)) as link:
+    with Handler(Framer(driver.read, driver.write, verbose=args.verbose)) as link:
         settings = Settings(link, timeout=args.timeout)
         with settings:
             if command == 'write':
--- a/piksi_tools/settings.py
+++ b/piksi_tools/settings.py
@@ -44,7 +44,6 @@ import configparser
 from collections import OrderedDict
 
 from sbp.client import Framer, Handler
-from sbp.logging import SBP_MSG_LOG
 from sbp.piksi import MsgReset
 from sbp.settings import (
     SBP_MSG_SETTINGS_READ_BY_INDEX_DONE, SBP_MSG_SETTINGS_READ_BY_INDEX_RESP,
@@ -55,12 +54,14 @@ from sbp.settings import (
 from piksi_tools import serial_link
 
 DEFAULT_READ_RETRIES = 5
-DEFAULT_CONFIRM_RETRIES = 2
-DEFAULT_WRITE_RETRIES = 5
 DEFAULT_TIMEOUT_SECS = 0.5
 
 KEY_ENCODING = 'ascii'    # encoding for settings sections and keys
 VALUE_ENCODING = 'ascii'  # encoding for settings values
+
+
+class ReadOnlySettingError(Exception):
+    pass
 
 
 class Settings(object):
@@ -77,7 +78,6 @@ class Settings(object):
                                SBP_MSG_SETTINGS_READ_BY_INDEX_RESP)
         self.link.add_callback(self._settings_done_callback,
                                SBP_MSG_SETTINGS_READ_BY_INDEX_DONE)
-        self.link.add_callback(self._print_callback, [SBP_MSG_LOG])
         return self
 
     def __exit__(self, *args):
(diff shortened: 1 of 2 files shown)


piksi_tools/settings.py
```diff
@@ -163,58 +159,63 @@ def read(self, section, setting, retries=DEFAULT_READ_RETRIES, verbose=False):
...
+        key = b'%s\0%s\0' % (section.encode(KEY_ENCODING), setting.encode(KEY_ENCODING))
+        request = key + b'%s\0' % value.encode(VALUE_ENCODING)
+
+        self.link(MsgSettingsWrite(setting=request))
+        reply = self.link.wait(SBP_MSG_SETTINGS_WRITE_RESP, timeout=self.timeout)
+
+        if reply is None:
+            raise RuntimeError("Settings write request {}/{} timed out"
+                               .format(section, setting))
+        # NOTE: reply payload is not set correctly when status is 2
```
pmiettinen: Check if the bug is still active with the latest firmware/buildroot master tip.

pmiettinen: If it is, assign the issue to me.

piksi_tools/settings.py
```diff
@@ -241,11 +242,11 @@ def write_from_file(self, output, verbose=False):
...
             for setting, value in settings.items():
-                return_code = self.write(section, setting, value, verbose=verbose)
-        return
-
-    def _print_callback(self, msg, **metadata):
-        print(msg.text.decode('ascii'))
+                try:
+                    self.write(section, setting, value, verbose=verbose)
+                except ReadOnlySettingError as e:
+                    print("  Warning: cannot write the read-only setting {}/{}, "
```
pmiettinen: You could just print `str(e)`?

piksi_tools/settings.py
tn-exafore: seems to work now, removing the workaround

piksi_tools/settings.py
tn-exafore: yeah, it's a valid simplification

tn-exafore:
note: the travis flake8 error comes due to a previous commit

piksi_tools/settings.py
```diff
@@ -77,7 +78,6 @@ def __enter__(self):
                                SBP_MSG_SETTINGS_READ_BY_INDEX_RESP)
         self.link.add_callback(self._settings_done_callback,
                                SBP_MSG_SETTINGS_READ_BY_INDEX_DONE)
-        self.link.add_callback(self._print_callback, [SBP_MSG_LOG])
```
denniszollo: removing the SBP_MSG_LOG callback may lose a human friendly textual hint from the firmware about why a settings write was rejected.  Do we think the return codes cover all the hints that are required?

piksi_tools/settings.py
```diff
@@ -266,50 +263,6 @@ def _settings_list_callback(self, sbp_msg, **metadata):
     def _settings_done_callback(self, sbp_msg, **metadata):
         self.settings_list_received = True
 
-    def _confirm_write(self, section, setting, value, retries, verbose=False):
```
denniszollo: removing the confirm write means that the tool will not support v1.0, v1.1, v1.2, or v1.3 of Piksi-Multi Firmware.  Is that acceptable?

piksi_tools/settings.py
tn-exafore: I think it works otherwise well except it lacks the "NTRIP must be disabled to modify settings" message, giving just a more generic "setting is read-only" error. However, this depends on new firmware.

tn-exafore:
Closed, but I'll keep the branch in my fork in case I'd want to return to this or take some aspects of this at some point in the future (not anytime soon at least).

PR #965, opened 2019-01-21
Add the verbose arg to Framer in settings and bootload_v3 [EX-476]

tn-exafore:


denniszollo:
I believe this means that we will see "host side unhandled byte" messages for traffic that is not SBP or partial messages.  I am okay with seeing that info for the bootload and settings so I approve.

tn-exafore:
Yeah, and also any messages about crc mismatch when the libsbp python3 PR eventually gets merged. I added these as a result of your comment in the libsbp WIP PR. The extra printouts would only appear when using the "-v" command line option for these cli tools.

PR #967, opened 2019-01-24
Update settings docs for IMU decimation [ESD-1014]

cbeighley:
Also update miscellaneous other out of date settings.

@silverjam @dzollo @nicolekelley

### DIFF
--- a/piksi_tools/console/settings.yaml
+++ b/piksi_tools/console/settings.yaml
@@ -2532,7 +2532,7 @@
   expert: true
   type: string
   units: N/A
-  default value: '23,65,72,74,81,97,117,134,136,137,138,139, 144,149,163,165,166,167,171,175,181, 185,187,188,189,190,257,258,259,520,522, 524,526,527,528,1025,2304/50,2305,2306/50, 30583,65280,65282,65535'
+  default value: '72,74,117,522,527'
   readonly: false
   Description: Configure which messages should be sent to the server.
   Notes: The enabled sbp messages settings is a list of message types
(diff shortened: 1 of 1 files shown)


piksi_tools/console/settings.yaml
```diff
@@ -2564,7 +2566,7 @@
   expert: true
   type: string
   units: N/A
-  default value: '72,74,117,522,527'
+  default value: '23,65,72,74,81,97,117,134,136,137,138,139,144,149,163,165,166,167,171,175,181,185,187,188,189,190,257,258,259,520,522,524,526,527,528,1025,2304,2305,2306,30583,65280,65282,65535'
```
silverjam: If these are the CAN ports (reviewing from mobile), I don’t think this is correct.  The CAN ports had an intentionally much more restrictive set of defaults.

piksi_tools/console/settings.yaml
```diff
@@ -824,7 +824,7 @@
   expert: false
   type: string
   units: N/A
-  default value: '23,65,72,74,81,97,117,134,136,137,138,139, 144,149,163,165,166,167,171,175,181, 185,187,188,189,190,257,258,259,520, 522,524,526,527,528,1025,2304,2305, 2306,30583,65280,65282,65535'
+  default value: '23,65,72,74,81,97,117,134,136,137,138,139,144,149,163,165,166,167,171,175,181,185,187,188,189,190,257,258,259,520,522,524,526,527,528,1025,2304/50,2305,2306/50,30583,65280,65282,65535'
```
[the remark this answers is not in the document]
cbeighley: Will add.

piksi_tools/console/settings.yaml
cbeighley: I took these from https://github.com/swift-nav/piksi_buildroot/blob/master/package/ports_daemon/ports_daemon/src/whitelists.c, I can just remove from this PR if you prefer.

silverjam: Ok, either way works, that's much more enabled messages than I expected it to have.

cbeighley: I'll remove for now.

PR #968, opened 2019-01-25
FileIO speed improvements [ESD-1038]

silverjam:
## Design notes
This PR adds several improvements to the Python `fileio` client, namely:

+ Bumps the size of the window we use to limit the number of outstanding packets 10x-- this is intended to make the window big enough that we should always have available space to allow us to always have request packets (for writing/reading) in-flight.  We want to avoid waiting in the `_check_pending` function for window space to become available as much as possible.

+  Tries to remove all loops from the fileio client.  Since we bumped the window size we're doing a lot more work per ACK packet that we receive.  To make the fileio client as fast as possible we remove all loops or `O(N)` ops on the number of outstanding requests.  The completion checking in particular now only looks at requests that are due to expire so that during a normal transfer (over TCP) the completion check should have to check zero packets for expiration.

## Testing

Tested on the bench using the `fileio.py` command line client and the console via `upgrade_view.py`.  Basic unit tests were added for the Time class in the `fileio` module.

### DIFF
--- a/piksi_tools/fileio.py
+++ b/piksi_tools/fileio.py
@@ -210,6 +210,8 @@ class SelectiveRepeater(object):
         self._expire_map = defaultdict(dict)
         self._callback_thread = None
         self._link_thread = None
+        self._total_sends = 1.0
+        self._total_retries = 0
 
     def __enter__(self):
         self._link.add_callback(self._request_cb, self._msg_type)
@@ -278,6 +280,8 @@ class SelectiveRepeater(object):
         Retry a request by updating it's expire time on the object
         itself and in the expiration map.
         """
+        self._total_retries += 1
+        self._total_sends += 1
         timeout_delta = Time(SBP_FILEIO_TIMEOUT)
         send_time = Time.now()
         new_expire = send_time + timeout_delta
@@ -331,6 +335,14 @@ class SelectiveRepeater(object):
             if not self._window_available(batch_size):
                 time.sleep(WAIT_SLEEP)
 
+    @property
+    def total_retries(self):
+        return self._total_retries
+
+    @property
+    def total_sends(self):
+        return self._total_sends
+
     def send(self, msg, batch_size=SBP_FILEIO_BATCH_SIZE):
         """
         Sends data via the current link, potentially batching it together.
@@ -351,6 +363,7 @@ class SelectiveRepeater(object):
             for msg in self._batch_msgs:
                 self._record_pending_req(msg, time_now, expiration_time)
             self._link(*self._batch_msgs)
+            self._total_sends += len(self._batch_msgs)
             del self._batch_msgs[:]
 
     def flush(self):
@@ -505,8 +518,8 @@ class FileIO(object):
                 sr.send(msg)
                 offset += len(chunk)
                 if (progress_cb is not None and seq % PROGRESS_CB_REDUCTION_FACTOR == 0):
-                    progress_cb(offset)
-            progress_cb(offset)
+                    progress_cb(offset, sr)
+            progress_cb(offset, sr)
             sr.flush()
 
 
@@ -651,7 +664,7 @@ def mk_progress_cb(file_length):
             previous_avg[0] = sum(rolling_avg_pts) / len(rolling_avg_pts)
             return previous_avg[0]
 
-    def the_callback(offset):
+    def the_callback(offset, repeater):
         time_current = time.time()
         offset_delta = offset - offset_last[0]
         time_delta = time_current - time_last[0]
@@ -659,8 +672,15 @@ def mk_progress_cb(file_length):
         mb_confirmed = offset / kb_to_mb
         speed_kbs = offset_delta / time_delta / 1024
         rolling_avg = compute_rolling_average(speed_kbs)
-        fmt_str = "\r[{:02.02f}% ({:.02f}/{:.02f} MB) at {:.02f} kB/s]"
-        status_str = fmt_str.format(percent_done, mb_confirmed, file_mb, rolling_avg)
+        fmt_str = "\r[{:02.02f}% ({:.02f}/{:.02f} MB) at {:.02f} kB/s ({:02.02f}% retries, {}/{})]"
+        percent_retried = repeater.total_retries / repeater.total_sends
+        status_str = fmt_str.format(percent_done,
+                                    mb_confirmed,
+                                    file_mb,
+                                    rolling_avg,
+                                    percent_retried,
+                                    repeater.total_retries,
+                                    repeater.total_sends)
         sys.stdout.write(status_str)
         sys.stdout.flush()
         time_last[0] = time_current
--- a/piksi_tools/fileio.py
+++ b/piksi_tools/fileio.py
@@ -39,8 +39,8 @@ SBP_FILEIO_WINDOW_SIZE = 1024
 #   at 20 bytes and TCP overhead at 20 bytes.
 SBP_FILEIO_BATCH_SIZE = 32
 
-SBP_FILEIO_TIMEOUT = 1
-MAXIMUM_RETRIES = 20
+SBP_FILEIO_TIMEOUT = 2
+MAXIMUM_RETRIES = 10
 PROGRESS_CB_REDUCTION_FACTOR = 100
 TEXT_ENCODING = 'utf-8'  # used for printing out directory listings and files
 
--- a/piksi_tools/fileio.py
+++ b/piksi_tools/fileio.py
@@ -155,22 +155,61 @@ class Time(object):
 
 
 class SelectiveRepeater(object):
-    """Selective repeater for SBP file I/O requests"""
+    """
+    Selective repeater for SBP file I/O requests
+
+    Fields
+    ----------
+    _pending_map : list(PendingRequest)
+      List (used as a map) of PendingRequest objects, used to track
+      outstanding requests.
+    _request_pool : Queue
+      Queue of available requests (recorded by index number)
+    _seqmap : dict(int,int)
+      Dictionary mapping SBP request sequence IDs to their corresponding
+      request index.
+    _batch_msgs : list
+      Collector for a batch of messages to be sent in one
+      buffer via the link
+    _last_check_time : Time
+      The last time we checked if any packets had expired
+    _expire_map : dict(Time, dict(PendingRequest, PendingRequest))
+      Dictionary which records the future time at which a request
+      will expire.
+
+    _msg_type : int
+      The message type we're currently sending
+    _link : Handler
+      The link over which we're sending data
+
+    _callback_thread : int
+      ID of the thread that we expect callbacks from
+    _link_thread : int
+      ID of the thread that handles link writes
+    """
 
     def __init__(self, link, msg_type, cb=None):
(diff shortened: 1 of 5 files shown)


piksi_tools/fileio.py
```diff
@@ -10,209 +11,416 @@
...
 MAX_PAYLOAD_SIZE = 255
 SBP_FILEIO_WINDOW_SIZE = 100
-SBP_FILEIO_TIMEOUT = 5.0
-MINIMUM_RETRIES = 3
+SBP_FILEIO_BATCH_SIZE = 1
+SBP_FILEIO_TIMEOUT = 3
+MAXIMUM_RETRIES = 20
 PROGRESS_CB_REDUCTION_FACTOR = 100
 TEXT_ENCODING = 'utf-8'  # used for printing out directory listings and files
+WAIT_SLEEP = 0.001
```
pmiettinen: `_S`?

piksi_tools/fileio.py
```diff
@@ -10,209 +11,416 @@
...
-        self.cb_thread = None
-        self.link_thread = None
+        """
+        Args
+        ---
+        link : Handler
+          Link over which messages will be sent.
+        msg_type:
+          The type of message being sent
+        """
```
pmiettinen: cb doc missing

piksi_tools/fileio.py
```diff
@@ -10,209 +11,416 @@
...
+        self._init_fileio_config(SBP_FILEIO_WINDOW_SIZE, SBP_FILEIO_BATCH_SIZE)
+
+        self._callback_thread = None
+        self._link_thread = None
+
+        self._total_sends = 1.0
+        self._total_retries = 0
+
+        self._config_send_time = Time.now()
+        self._config_retry_time = Time.now()
```
pmiettinen: Should these be initialized to same instance/moment of time?

piksi_tools/fileio.py
```diff
@@ -351,12 +559,13 @@ def write(self, filename, data, offset=0, trunc=True, progress_cb=None):
                 msg = MsgFileioWriteReq(
                     sequence=seq,
                     offset=offset,
-                    filename=filename + b'\x00',
-                    data=chunk)
+                    filename=filename + b'\x00' + chunk,
```
pmiettinen: Why like this?

benjaminaltieri: This one deserves an entire post-mortem on it own.

silverjam: I'll add a comment, but basically it's 2 or 3 times faster (FileIO itself) if we use the "wrong" form of this.

requirements.txt
```diff
@@ -12,7 +12,7 @@ pyparsing==2.2.0
 pygments==2.2.0
 requests==2.20.0
 six==1.10.0
-sbp==2.4.5
+git+https://github.com/swift-nav/libsbp@4e8fc16b95639cd58649fe2d559230798f714e65#egg=sbp&subdirectory=python
```
denniszollo: Reminder to replace this with a tag when the requisite libsbp versions is released.

piksi_tools/fileio.py
```diff
@@ -10,209 +11,416 @@
...
 from piksi_tools import serial_link
 
 MAX_PAYLOAD_SIZE = 255
 SBP_FILEIO_WINDOW_SIZE = 100
-SBP_FILEIO_TIMEOUT = 5.0
-MINIMUM_RETRIES = 3
+SBP_FILEIO_BATCH_SIZE = 1
+SBP_FILEIO_TIMEOUT = 3
+MAXIMUM_RETRIES = 20
 PROGRESS_CB_REDUCTION_FACTOR = 100
```
denniszollo: Should PROGRESS_CB_REDUCTION_FACTOR change based upon the rate at which FILEIO is operating?  I tuned this based upon the speed of the old file transfer, but now if we move 3 times faster, I think perhaps this factor should be reduced accordingly when we receive the config response.  Basically I saw that updating the GUI was eating CPU and making a crash of the console more likely.

piksi_tools/fileio.py
```diff
@@ -10,209 +11,416 @@
...
-        `self.write_pool`.
+        Return a pending request to the write pool and clean any
+        entries in the expiration map.
         """
         self._verify_cb_thread()
-        self.pending.pop(self.pending.index(pending_write))
-        self.write_pool.append(pending_write.invalidate())
+        self._request_pool.put(pending_req)
+        del self._seqmap[pending_req.message.sequence]
+        del self._expire_map[pending_req.time_expire][pending_req]
```
benjaminaltieri: when testing with vanilla 1.5/2.0/2.1/2.2 I'm getting quite a few of these:
```
  File "piksi_tools/fileio.py", line 262, in _return_pending_req
    del self._expire_map[pending_req.time_expire][pending_req]
  KeyError: PendingRequest(offset=18924416,seq=833422368,time=Time(s=1549918415,ms=12),tries=0,index=58)
```
although transfers seem to complete (although I'm testing with a junk file, verification is not being tested). maybe the `del` needs to be wrapped with some other logic?

benjaminaltieri: Actually I just got stuck on a transfer with 2.2 but it's really hard to tell if this was a pre-existing condition:
screen shot 2019-02-11 at 1 41 02 pm
May not be actionable for this PR but wanted to note since I took the time to check these out.

piksi_tools/fileio.py
```diff
@@ -474,7 +683,60 @@ def printable_text_from_device(data):
     return str
 
 
+def mk_progress_cb(file_length):
+
+    time_last = [time.time()]
```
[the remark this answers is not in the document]
silverjam: Since this is for the progress function we don't need to use `Time.now()` -- `Time.now()` is only special in that it only tracks seconds and milliseconds.

piksi_tools/fileio.py
silverjam: Not sure how this could happen, unless we somehow lost a chunk of the file that we thought timed out and never requested a retry.

piksi_tools/fileio.py
```diff
@@ -10,209 +11,416 @@
...
+
+        self._total_sends = 1.0
+        self._total_retries = 0
+
+        self._config_send_time = Time.now()
+        self._config_retry_time = Time.now()
+        self._config_seq = random.randint(0, 0xffffffff)
+        self._config_msg = None
+
+        self._link(MsgFileioConfigReq(sequence=self._config_seq))
```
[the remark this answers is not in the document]
silverjam: Can probably move this into `self._wait_config_received` so we can do this as part of the start of the first send operation.

piksi_tools/fileio.py
silverjam: Pretty sure this is OK just because of the order things are done in `self._config_receive()` but it probably makes more sense to setup the next expire time here and then also in the assignment of the value here https://github.com/swift-nav/piksi_tools/pull/968/files#diff-619782351bcba823c8db86598d3ca1b6R355.

piksi_tools/fileio.py
silverjam: ok, I can experiment with that

requirements.txt
silverjam: 🔨

piksi_tools/fileio.py
silverjam: 🔨

piksi_tools/fileio.py
silverjam: 🔨

piksi_tools/fileio.py
silverjam: Fixed

piksi_tools/fileio.py
silverjam: Added some code to double PROGRESS_CB_REDUCTION_FACTOR if we're in the "new" config

piksi_tools/fileio.py
```diff
@@ -474,7 +683,60 @@ def printable_text_from_device(data):
     return str
 
 
+def mk_progress_cb(file_length):
+
+    time_last = [time.time()]
+    offset_last = [0]
+
+    kb_to_mb = 1024 * 1024.0
```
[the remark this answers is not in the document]
silverjam: 🔨

piksi_tools/fileio.py
silverjam: I think I found the issue (I was able to reproduce the hang by running transfers in an infinite loop), there's a small change that we could "return" a pending request that was already returned (not exactly sure how)-- but fixed here 9a797c2b251772745fbe0bae0eeaedadf2fb28d1.  Ran a transfer in a loop overnight and didn't see the hang happen again.

piksi_tools/fileio.py
silverjam: I'm still seeing a lot of "We dropped a packet, skipping this observation sequence..." logs, but I am not convinced it's the console that's dropping things... I think in this high bandwidth scenario we might be dropping data that's coming out of rpmsg somehow.  The dropped packet logging gets much worse if I up the solution rate.

piksi_tools/fileio.py
silverjam: I added a comment about this, but also experimented with some other options, some of which resulted in pretty large speed-ups (30% increase in max transfer speed) such as this one: https://github.com/swift-nav/piksi_tools/commit/0ebf8d91d7870c30e5a51c142427b58f985b313f -- which basically adds numba (JIT compiler for individual Python functions) to speed up the CRC calculation and also hand rolled the write message packet.  There's a lot of other work that could probably be done too, like trying to avoid copies in that pipeline, but wasn't able to get that work yesterday.

piksi_tools/fileio.py
silverjam: Also, check out e1139f4c6d9d8deea60b3f587906995ee1e3b5f0 -- which adjust the logic around this message, before this change I was seeing that we always logged twice when we got into this situation.

piksi_tools/fileio.py
```diff
@@ -286,7 +494,7 @@ def readdir(self, dirname=b'.'):
             msg = MsgFileioReadDirReq(
                 sequence=seq, offset=len(files), dirname=dirname)
             self.link(msg)
-            reply = self.link.wait(SBP_MSG_FILEIO_READ_DIR_RESP, timeout=1.0)
+            reply = self.link.wait(SBP_MSG_FILEIO_READ_DIR_RESP, timeout=5.0)
```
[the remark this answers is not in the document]
silverjam: 🔨

piksi_tools/fileio.py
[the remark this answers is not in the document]
silverjam: 🔨

piksi_tools/fileio.py
```diff
@@ -10,209 +11,416 @@
...
-    def track(self, pending_write, time, tries=0, complete=False):
+    def track(self, pending_req, time, time_expire, tries=0):
         """
         Load information about the pending write so that it can be tracked.
         """
-        self.message = pending_write
+        self.message = pending_req
         self.time = time
+        self.time_expire = time_expire
         self.tries = 0
```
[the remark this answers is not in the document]
silverjam: 🔨

piksi_tools/fileio.py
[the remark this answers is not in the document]
silverjam: 🔨

PR #970, opened 2019-01-30
Handle failed connections gracefully [ESD-1034]

benjaminaltieri:
This PR refactors the console modules startup process to clearly delineate connection attempts from the port chooser dialog. It also centralizes all driver selection around the `serial_link.get_base_args_driver` method.

About 50% of the changes are simple refactoring around that central method for establishing connections based on args from cli or port chooser. The other half is adding the dialog that indicates what may have gone wrong in the connection and falls back to port chooser to allow the user to try again as opposed to restarting the program completely. A typical flow would look like this:

  - User selects some command arguments via CLI
  - Connection fails, a dialog is shown
  - Selecting 'OK' will move to port chooser dialog for retry, 'Cancel' or closing the window will exit the program
  - This will loop back until a successful connection is made, or the user cancels the attempts

### DIFF
--- a/piksi_tools/console/console.py
+++ b/piksi_tools/console/console.py
@@ -30,17 +30,18 @@ from sbp.navigation import SBP_MSG_POS_LLH
 from sbp.piksi import SBP_MSG_COMMAND_RESP, MsgCommandResp, MsgReset
 from sbp.system import SBP_MSG_HEARTBEAT
 from traits.api import (Bool, Dict, Directory, Enum, HasTraits, Instance, Int,
-                        List, Str)
+                        Str)
 # Toolkit
 from traits.etsconfig.api import ETSConfig
-from traitsui.api import (EnumEditor, Handler, HGroup, HTMLEditor, ImageEditor,
-                          InstanceEditor, Item, Label, Spring,
+from traitsui.api import (Handler, HGroup, HTMLEditor, ImageEditor,
+                          InstanceEditor, Item, Spring,
                           Tabbed, TextEditor, UItem, VGroup, View, VSplit)
 
 import piksi_tools.serial_link as s
 from piksi_tools import __version__ as CONSOLE_VERSION
 from piksi_tools.console.baseline_view import BaselineView
 from piksi_tools.console.callback_prompt import CallbackPrompt, ok_button
+from piksi_tools.console.port_chooser import get_args_from_port_chooser
 from piksi_tools.console.deprecated import DeprecatedMessageHandler
 from piksi_tools.console.imu_view import IMUView
 from piksi_tools.console.mag_view import MagView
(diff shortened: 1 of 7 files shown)


piksi_tools/console/console.py
```diff
@@ -787,128 +786,131 @@ def __init__(self, usage, error_str):
...
+        else:
+            cnx_data = ConnectionData(
+                driver=driver,
+                description=description,
+                mode=mode
+            )
+    return (cnx_data, connect_error)
+
+
+def port_chooser_retry_logic(args):
```
pmiettinen: Should theese be in the `port_chooser.py`?

piksi_tools/utils.py
```diff
@@ -58,4 +58,5 @@ def get_tcp_driver(host, port=None):
         raise Exception('TCP connection timed out. Check host: {}'.format(host))
     except Exception as e:
         import traceback
-        raise Exception('Invalid host and/or port: {0}'.format(traceback.format_exc()))
+        print('Invalid host and/or port: {0}'.format(traceback.format_exc()))
+        raise Exception('Invalid host and/or port')
```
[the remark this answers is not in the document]
benjaminaltieri: done. and removed random print statement

piksi_tools/console/console.py
benjaminaltieri: I left it out as it uses `connection_from_args`, basically it the how-console-uses-port-chooser method

piksi_tools/console/console.py
```diff
@@ -834,24 +834,23 @@ def set_connection_mode(self, mode):
...
     connect_error = None
-    retry_connection = True
     description = ''
     mode = ''
     port = getattr(args, 'port', None)
     baud = getattr(args, 'baud', None)
 
     if port is not None:
-        if args.tcp:
+        if args.tcp is True:
```
silverjam: Sorry, should have explained this more, the `is None` and `is not None` idiom usually only applies to the None value (which is a singleton in Python) and is basically done because it avoids any equality operators that an object might implement and forces Python to evaluate the question "is this object identical to the None singleton" instead of "call the equality operator on this object and pass None".

So, I may've given misleading advice on the idiom early... but if you want to check the "truthiness" of something, just using `if blah` is fine (None, False, empty lists/sets will all evaluate to false).  However, if you explicitly want to know if something is undefined/None (for example, to initialize the value because the argument wasn't specified) then `is None` or `is not None` is more appropriate.

benjaminaltieri: gotcha, basically I am using `getattr` to avoid exceptions (as opposed to using a `try except` for each) if `args` doesn't have the common set of values defined. I'm sure this is non-idiomatic as well *shrug*

PR #977, opened 2019-02-12
Allow AppVeyor/PyInstaller to use libsbp from Git [ESD-1038]

silverjam:
This changes make `piksi_tools` tolerant of a `libsbp` version that is fetched from a Git repository by commit hash rather than PyPI-- this change allows our CI infrastructure to pick up development versions of libsbp that are paired with development work in the console.


silverjam:
( cc @tn-exafore @pmiettinen )

PR #978, opened 2019-02-13
Adding Console logic to enable Console Dynamic Velocity Profiles capa…

arunabh-c:
…bility

New option added to console script to enable process noise motion modes toggle

### DIFF
--- a/piksi_tools/console/settings.yaml
+++ b/piksi_tools/console/settings.yaml
@@ -405,6 +405,21 @@
     attempted by the receiver.
 
 - group: solution
+  name: process_noise_motion_mode
+  expert: false
+  type: enum
+  units: N/A
+  default value: 'Default'
+  readonly: false
+  enumerated possible values: Default, Automotive, Static
+  Description: Selects the process noise variance of position, velocity & acceleration in the Horizontal &
+    Vertical directions.
+  Notes: The "Default" mode assigns equal process noise variances (uncertainities in the dynamic model) in all
+    directions. The "Automotive" mode assigns relatively smaller process noise variances to the vertical
+    position/velocity/acceleration. The "Static" mode assigns small process noise variances for position/velocity
+    /acceleration in both the horizontal and vertical directions.
+
+- group: solution
   name: dgnss_filter
   expert: true
   type: enum
--- a/piksi_tools/console/settings.yaml
+++ b/piksi_tools/console/settings.yaml
@@ -406,13 +406,13 @@
 
 - group: solution
   name: process_noise_motion_mode
-  expert: false
+  expert: true
   type: enum
   units: N/A
   default value: 'Default'
   readonly: false
   enumerated possible values: Default, Automotive, Static
-  Description: Selects the process noise variance of position, velocity & acceleration in the Horizontal &
+  Description: Selects the Filter Uncertainity of position, velocity & acceleration in the Horizontal &
     Vertical directions.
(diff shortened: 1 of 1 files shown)


arunabh-c:
Corresponding Starling PR: https://github.com/swift-nav/starling/pull/1975
Corresponding PFWP PR: https://github.com/swift-nav/piksi_firmware_private/pull/2719

piksi_tools/console/settings.yaml
```diff
@@ -404,6 +404,21 @@
...
     attempted by the receiver.
 
+- group: solution
+  name: process_noise_motion_mode
+  expert: false
+  type: enum
+  units: N/A
+  default value: 'Default'
+  readonly: false
+  enumerated possible values: Default, Automotive, Static
```
pgrgich: As for the other PR, I'm a bit unsure what our options should be. If we want to keep the current configuration with high process noise, we should probably just name it something that makes it obvious what it is (or perhaps "UAV", if that's what the UAV use case will use). It sounds like "Static" has changed a bit, but why does the static case have non-zero process noise? Is it assumed that tectonic motion needs to be considered? If it's more than that, I'd think the name should be changed to "Low dynamics" or something.

anth-cole:
Can we make sure this is coordinates with Alex? This is the customer facing piece so we should get product input.

I'm not a fan of the setting name, it requires the user to know what the process noise is, what about 'dynamic motion model' or something?

piksi_tools/console/settings.yaml
```diff
@@ -404,6 +404,21 @@
     for each timestamp. When "No DGNSS" is chosen, no differential output will be
     attempted by the receiver.
 
+- group: solution
+  name: process_noise_motion_mode
```
scarcanague: Have you discussed with Alex on this? `dynamic motion model'` sounds indeed more customer friendly

piksi_tools/console/settings.yaml
```diff
@@ -404,6 +404,21 @@
...
+  expert: false
```
scarcanague: That should be true, I dont think we want the setting in the default intially

piksi_tools/console/settings.yaml
```diff
@@ -404,6 +404,21 @@
...
+  type: enum
+  units: N/A
+  default value: 'Default'
+  readonly: false
+  enumerated possible values: Default, Automotive, Static
+  Description: Selects the process noise variance of position, velocity & acceleration in the Horizontal &
```
scarcanague: We probably also want something a little more customer friendly for the description

arunabh-c:
> As with the other PRs, I'm not sure what the status is for naming. Can you confer with Sebastien, Alex and anyone else they consider needs to be involved in order to get to a conclusion?

waiting to hear back from alex on final version for names and description.

pgrgich:
For the purposes of our first cut, shall we go with something like:
Model name: `dynamic motion model`
Model options: `HIGH_DYNAMICS`, `HIGH_HORIZONTAL_DYNAMICS` and `LOW_DYNAMICS` (replacing `DEFAULT`, AUTOMOTIVE` and `STATIC`).
Model descriptions: `High dynamics - high process noise in all axes`, `High horizontal dynamics - high process noise in horizontal plane and low process noise in vertical axis` and `Low dynamics - low process noise in all axes`
Happy for this to be changed, but I think we need something to iterate on.

PR #979, opened 2019-02-13
Ignore update warning

denniszollo:
WIP attempt to allow arbitrary fw upgrade in console but still warn users if they might have a v2.0.0 requirement before version jumping.

PR #980, opened 2019-02-15
Can't upgrade to or from develop version [ESD-1083]

woodfell:
# Goal
The root cause is that firmware builds use version tags generated by git, but console is using a python version string standard parsing library. Development version strings fail to parse and fall fowl the logic to manage the convoluted upgrade path between < v2.0.0 to >= 2.0.0

# Testing

Development versions should be accepted in all upgrade situations, no matter what the currently installed version is

Verify upgrade path 1.4.* -> 2.0.0 -> 2.2.17

Verify upgrade path 1.4.* -> 2.2.17 is blocked, console prompt to upgrade to v2.0.0 first


piksi_tools/console/GitVersion.py
```diff
@@ -0,0 +1,108 @@
...
+        return self._version
+
+    def __repr__(self):
+        return "<GitVersion({0})>".format(self._version)
+
+    def compare_dev_string(self, other):
+        return self.devstring.compare(other.devstring)
+
+    def _is_dev(self):
+        return len(self.devstring) > 0 and self.devstring != 'v'
```
benjaminaltieri: quick comment here on why 'v' is a special case could be helpeful

piksi_tools/console/GitVersion.py
```diff
@@ -0,0 +1,108 @@
...
+    @property
+    def devstring(self):
+        return self._devstring
+
+    @property
+    def isdev(self):
+        return self._is_dev()
+
+    _regex = re.compile(
+        r"^\s*(?P<leader>[^0-9]*)(?P<marketing>[0-9]+)\.(?P<major>[0-9]+)\.(?P<minor>[0-9]+)(?P<dev>.*)$",
```
benjaminaltieri: this is great. a docstring with a couple examples of current version strings and how they would fill out the match groups could help de-obfuscate the expected behavior for anyone not regex-literate, even though it's pretty clear on it's own.

benjaminaltieri:
great stuff @woodfell - added a couple random comment but overall this is super clean Python OO. When you have a chance, fill in the detail in the PR description on the aim and strategy for testing before you hit squash.

denniszollo:
Can we add some unit tests to this like I did on my attempt?
https://github.com/swift-nav/piksi_tools/pull/979

In this way there is proof in the pudding that it will handle all versions strings correctly, including those from Swift that don't conform to any PEP stuff.

benjaminaltieri:
Agree, tests would be good to have a record of the sorts of strings we've evaluated in the past.

denniszollo:
The key strings to test and evaluate look like this:
v2.1.0-develop-2019021521-4-gee0b78d4

The regex will fail on this in looking by inspection.

benjaminaltieri:
Not sure about that. I would expect it to parse with the following groups:

leader: v
marketing: 2
major: 1
minor: 0
dev: -develop-2019021521-4-gee0b78d4

woodfell:
Added unit tests as suggested.

It isn't right for console to be using the PEP spec, at least when dealing with piksi firmware images, since those version numbers are generated by git and are never going to be compliant with any python specification.

The way we handle, parse, and group the sections of an input string is pretty weird. For example, any digits which are in the leader invalidate the string. Leading whitespace is stripped, but trailing whitespace isn't. And the leader and dev groups are combined in to the 'devstring' property. This behaviour is appropriate because it mirrors the behaviour of libpiksi, but I question whether it is the best way to handle version strings. I will write up a proposal to address these points.

Additionally, console uses the firmware image filename to generate the 'intended' version. This means the console can be defeated simply by renaming the file to upload. A fix for this is not included in this bug.

benjaminaltieri:
:shipit:

PR #981, opened 2019-02-15
Adapt console fw update to fileio python 3 changes [EX-465]

tn-exafore:
Console python 3 support is a WIP, this just updates the fw update view that was broken by the cli tools python 3 change.


tn-exafore:
woodfell merged this change as part of #980 so this PR is not needed anymore; closing

PR #984, opened 2019-02-25
Unicode username fix

denniszollo:
This PR attempts to fix an issue where the users home directory has unicode characters in the path reported on windows 10.

### DIFF
--- a/piksi_tools/console/utils.py
+++ b/piksi_tools/console/utils.py
@@ -5,6 +5,7 @@ import datetime
 import pkg_resources
 import time
 import os
+import locale
 
 from functools import partial
 from threading import Event, Thread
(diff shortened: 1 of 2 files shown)


piksi_tools/console/utils.py
```diff
@@ -638,4 +639,9 @@ def loop():
 resource_stream = partial(pkg_resources.resource_stream, 'piksi_tools')
 icon = ImageResource(resource_filename('console/images/icon.png'))
 
-swift_path = os.path.normpath(os.path.join(os.path.expanduser("~"), 'SwiftNav'))
+home_path = os.path.expanduser("~" + os.sep)
+loc = locale.getdefaultlocale()
```
benjaminaltieri: storing the result as the decomposed tuple might be more clear here ([docs](https://docs.python.org/2/library/locale.html#locale.getdefaultlocale)):
```suggestion
(language_code, encoding) = locale.getdefaultlocale()
```
and explicitly testing for `None` to indicate that `encoding` is not a boolean

denniszollo:
thanks ben, forced push that suggestion on-top of master.  When CI is green I will plan to merge.

silverjam:
All green! Let's merge! 🚀

PR #989, opened 2019-03-14
Add option to omit metadata from JSON log [ESD-1170]

pmiettinen:
https://github.com/swift-nav/libsbp/pull/663

With the new option (similar to haskell `sbp2json` but without compact encoding):
`{"sender": 45490, "msg_type": 528, "age": 65535, "tow": 800, "crc": 40222, "length": 6, "preamble": 85, "payload": "IAMAAP//"}`

Without the new option:
`{"data": {"sender": 45490, "msg_type": 528, "age": 65535, "tow": 800, "crc": 40222, "length": 6, "preamble": 85, "payload": "IAMAAP//"}, "session-uid": "f3106e95-2dc0-4e3d-97ff-1474c7123a1b", "time": "2019-03-14T08:44:02.446149Z"}`

### DIFF
--- a/piksi_tools/serial_link.py
+++ b/piksi_tools/serial_link.py
@@ -123,6 +123,10 @@ def base_cl_options(override_arg_parse=None, add_help=True,
             default=False,
             help="Expand fields in JSON logs"
         )
+        parser.add_argument(
+            "--skip-metadata",
+            action="store_true",
+            help="Omit metadata from JSON logs.")
     return parser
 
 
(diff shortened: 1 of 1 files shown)


piksi_tools/serial_link.py
```diff
@@ -95,6 +95,11 @@ def base_cl_options(override_arg_parse=None, add_help=True,
         "--verbose",
         action="store_true",
         help="print extra debugging information.")
+    parser.add_argument(
+        "-m",
+        "--metadata-skip",
```
silverjam: How about `--skip-metadata`

pmiettinen: That's probably better. Also I think the short switch can be left out.

PR #990, opened 2019-03-14
"Thread died!" error at eof [ESD-1167]

pmiettinen:
https://swift-nav.atlassian.net/browse/ESD-1167

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -79,3 +79,6 @@ swift_nap_*.hex
 
 # Virtual env
 *venv
+
+# IDE files
+.vscode/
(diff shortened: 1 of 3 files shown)


pmiettinen:
@silverjam Maybe behind verbose?

pmiettinen:
@silverjam I created a separate issue for the `sys.exit` removal. Seems like the comment for the `KeyboardInterrupt` is outdated etc. so needs some further investigations what is supposed to happen on each execution path.

https://swift-nav.atlassian.net/projects/ESD/issues/ESD-1187

PR #994, opened 2019-03-22
Update linux python3 platform to use pyqt5 [EX-468][EX-537]

tn-exafore:
* fixes a memory leak in tables
* use pyqt5 v. 5.10.0 because 
  * 5.12 has an UI bug https://bugreports.qt.io/browse/QTBUG-73721
  * pyinstaller 3.4 does not support 5.11
  * v. 5.10.1 has issues with qtwebengine
* pyqt5 is only available on pypi for python 3 so mac & win, whose build configs are still for python 2.7, will be updated later on
* switch the legacy linux python 2.7 platform from pyqt4 back to pyside because it's available on pypi and thus simpler to use with venvs and tox


silverjam:
Sorry if it's been said before, but what's the rationale for moving only Linux to py3 and pyqt5? What are the remaining blockers for Mac and Windows?

tn-exafore:
> 
> 
> Sorry if it's been said before, but what's the rationale for moving only Linux to py3 and pyqt5? What are the remaining blockers for Mac and Windows?

The rationale is one step at a time. There are probably some build changes needed for Win and Mac that are specific to those platforms, and this way we can get something in faster.

tn-exafore:
FYI: if somebody wants to do a test-run of this PR, there is an unrelated python 3 problem currently in master as commented in https://swift-nav.atlassian.net/browse/ESD-1113

PR #995, opened 2019-03-25
setuptools_scm version output now includes 'v' so stop adding it [ESD-1195]

benjaminaltieri:
~WIP - testing build output artifacts for now~

seems to do the trick. moving forward I believe the assumption should be that the `v` in `v2.1.17` etc., is part of the `CONSOLE_VERSION` that comes from `_version.py`. I don't know exactly when/why this changed but I think a missing `v` is better than `vv` so this PR simply errs on the side of removing all artificially added `v`'s

### DIFF
--- a/piksi_tools/console/console.py
+++ b/piksi_tools/console/console.py
@@ -113,7 +113,7 @@ def get_args():
     return parser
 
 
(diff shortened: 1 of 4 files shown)


benjaminaltieri:
OSX builds keep timing out for some reason, unrelated to this PR

PR #998, opened 2019-03-28
Update RAIM Disable decription

anth-cole:
Disabling RAIM only turns it off on the ME (controlling observation output) and not in starling today


pgrgich:
Lgtm.

PR #1000, opened 2019-03-29
Bump to libsbp 2.5.2, use Conda for OS X [ESD-1208]

silverjam:
- Use Conda managing Python environments on OS X, this greatly simplifies the build environment setup and allows us to align on Python 3.5 for Linux/Windows.

- We need to use our fork of enable (now based on upstream with the damage memory leak fixes) because there's issues building on OS X otherwise... we should push these fixes upstream as soon as possible.

- Bump libsbp to version 2.5.1, conda seems to resolve the build issue we were seeing here because wheels for numba/numpy "just work" under the conda generated Python environments.


pmiettinen:
@silverjam - I was thinking the same that probably the whole OSX setup process needs to be overhauled. Especially `brew install python@2` seems way too intrusive for the user's environment.

silverjam:
@pmiettinen @tn-exafore Got kind of carried away on this, sorry, conda seemed to magically solve most of the setup issues on OS X ... wondering if it shouldn't be extended to the rest of the platforms (or, at least, for the GUI builds of things).

pmiettinen:
I am using conda in libsettings CI with windows builds. Seems to be easy enough way to handle the different Python versions. There's some extra twitch due to the travis using git bash on the windows containers but nothing that I haven't been able to overcome with brute force trial and error this far. No experience with conda on unix systems though. On OSX/Ubuntu I have been using pyenv instead at it's been working great switching between versions.

silverjam:
@pmiettinen conda seemed more appropriate since it manages non-Python things as well, like swig and the libraries we need to compile things since we don't yet have wheels for enable/traits

tasks/setup.sh
```diff
@@ -170,66 +172,59 @@ function piksi_splash_osx () {
...
+    pip install -r "$ROOT/requirements_dev.txt"
+    pip install -r "$ROOT/requirements.txt"
+    pip install -r "$ROOT/requirements_gui.txt"
+    pip install -e "$ROOT"
+
+    pip install PyQt5==5.10.0
+
+    log_info ""
+    log_info "To run piksi_tools from source, do the following:"
+    log_info "  source activate ${conda_env_name}"
```
pmiettinen: If done right after `make deps` does this instruction lead to possible double activation?

silverjam: If you run activate afterward, then run `make deps` again it'll cause an error with a double activation, but the activate earlier in this script it local to the `setup.sh` script so it doesn't affect the calling environment.

pmiettinen: Ok, maybe add deactivate info similar what is printed after creating conda env?

silverjam: Sure, will add

tox.ini
```diff
@@ -20,10 +22,12 @@ commands =
 [testenv:gui35]
 basepython=python3.5
 passenv = TOXENV DISPLAY HOME
+whitelist_externals = bash py.test
 commands =  
            # hack since enable is not packaged correctly
            pip install -r{toxinidir}/requirements_gui.txt
            pip install pyqt5==5.10.0
+           bash {toxinidir}/scripts/font_manager.bash
```
tn-exafore: what kind of an issue is this hack for?

requirements_dev.txt
```diff
@@ -0,0 +1,4 @@
+cython
```
tn-exafore: should versioning be added for cython?

tox.ini
silverjam: font manager spews a bunch of exceptions on start-up while it's building it's cache, it doesn't look like it's really an error, but it creates a lot of noisy logging, this squelches it

requirements_dev.txt
silverjam: I don't think we need it, as far as I know, Cython generates code that's backwards compatible with all Python versions, so we shouldn't pin the version until we have a reason to

tox.ini
```diff
@@ -4,7 +4,9 @@ minversion = 2.0
 
 [testenv]
 deps = -r{toxinidir}/requirements.txt
+       -r{toxinidir}/requirements_dev.txt
 commands = py.test -v --ignore=tests/test_console.py tests/
+whitelist_externals = bash py.test
```
tn-exafore: isn't `py.test` implicitly supported by tox without whitelisting?

requirements_dev.txt
tn-exafore: maybe a minimum version >=0.28?

tox.ini
tn-exafore: maybe add a comment about it?

tox.ini
silverjam: will fix

tasks/setup.sh
silverjam: Fixed

requirements_dev.txt
silverjam: Added min version

tox.ini
silverjam: Good idea, added comments

.travis.yml
```diff
@@ -44,30 +44,27 @@ matrix:
...
      env: TESTENV=build_console
-     before_install: make deps
+     before_install:
+       - make deps
+       - pip install -r requirements_dev.txt
      script: make build_console
 
    - os: osx
      osx_image: Xcode 8.3.3
-     env: TESTENV=mac TOXENV=py27,py37
```
tn-exafore: This `TESTENV=mac` had so far been just an informational variable for travis. IMO it would be clearer to define at least that and probably also `TOXENV` (if the latter doesn't cause difficulties, haven't yet reviewed all) in .travis.yml instead of in Makefile, but this is just an opinion.

.travis.yml
```diff
@@ -44,30 +44,27 @@ matrix:
...
      language: generic
      before_install:
-        - export PATH="/usr/local/opt/python@2/libexec/bin:$PATH"
+        - export PATH="/Users/travis/miniconda3/bin:$PATH"
         - make deps
-     before_script:
-        - export PATH="/usr/local/bin/:./:$PATH"
-        - export DISPLAY=:0.0
-        - Xvfb ${DISPLAY} -screen 0 800x600x16 & PID=$!
```
tn-exafore: doesn't this break running the gui smoketest on mac?

.travis.yml
silverjam: fixed

.travis.yml
silverjam: As far as I know, Xvfb is a X11 windows display manager (and we're not using X11 on Mac), should be completely unnecessary for the Mac build and the smoke tests seem to run fine without it.

.travis.yml
```diff
@@ -44,30 +44,28 @@ matrix:
...
+     before_install:
+       - make deps
+       - pip install -r requirements_dev.txt
      script: make build_console
 
    - os: osx
      osx_image: Xcode 8.3.3
-     env: TESTENV=mac TOXENV=py27,py37
      language: generic
+     env: TESTENV=mac TOXENV=py27,gui35
```
tn-exafore: I think you should add also either py35, py37 or both here so that the cli tests would still get executed on mac for python 3, too.

silverjam: 👍

.travis.yml
tn-exafore: ok, probably there is some "display" available in travis' mac environment by default, then. Smoke test running without any special errors/warnings is enough.

.travis.yml
```diff
@@ -44,30 +44,28 @@ matrix:
...
-        - export PATH="/usr/local/opt/python@2/libexec/bin:$PATH"
+        - export PATH="/Users/travis/miniconda3/bin:$PATH"
         - make deps
-     before_script:
-        - export PATH="/usr/local/bin/:./:$PATH"
-        - export DISPLAY=:0.0
-        - Xvfb ${DISPLAY} -screen 0 800x600x16 & PID=$!
-        - sleep 3  # give xvfb some time to start
      script:
-        - tox -v
```
tn-exafore: This is a heavily subjective opinion, feel free to ignore, but wouldn't it be a bit simpler to do just the conda setup in the Makefile, and call that target from here, and then tox directly from here? Now there are several targets in Makefile just for testing...

Makefile
```diff
@@ -43,15 +34,45 @@ all: deps
...
+tox_all:
+	@echo Using TESTENV=$(TESTENV), TOXENV=$(TOXENV)...
+	tox $(if $(filter y,$(VERBOSE)), -v,)
+
+tox_Darwin: export TESTENV:=$(TESTENV)
+tox_Darwin: export TOXENV:=$(TOXENV)
+tox_Darwin: tox_all
+
+tox: .conda_py27 .conda_py35
+tox: export PATH:=$(CURDIR)/.conda_py35/bin:$(CURDIR).conda_py27/bin:$(PATH)
```
tn-exafore: would the path expansions be neater in .conda_py27 and ._conda_py35 targets?

.travis.yml
silverjam: Reintroducing Python 3.7 will probably bring back the original CI problem, will let Travis bear this out and check on things in the morning.

tasks/setup.sh
```diff
@@ -20,6 +20,8 @@
 ####################################################################
 ## Utilities.
 
+ROOT=$( (cd "$(dirname "$0")/.." || exit 1 >/dev/null; pwd -P) )
```
tn-exafore: Did the earlier relative paths fail with anaconda? Do you envision global use for this or would it be better in install_python_deps_osx()?

.travis.yml
silverjam: I feel like the relationship between `.travis.yml`, the `Makefile`, and `tasks/setups.sh` are a bit ill-defined / muddy.  In general I'd to most data not in Travis since it makes testing much easier and makes the CI more portable to other systems (if we ever need to migrate).  I tested many of these things on a Mac OS X 10.13 VM.

Makefile
silverjam: Pretty sure each env variable that's added is only present within that target invocation, so this would look cleaner, but I don't think it would work.

Example:
```shell
$ cat Makefile
all: a

b: export B_VAR="b-var-value"
b:
        @echo "In 'b' target..."
        @echo B_VAR=$(B_VAR)

a: export A_VAR="a-var-value"
a: b
        @echo "In 'a' target..."
        @echo A_VAR=$(A_VAR)
        @echo B_VAR=$(B_VAR)

$ make
In 'b' target...
B_VAR=b-var-value
In 'a' target...
A_VAR=a-var-value
B_VAR=
```

tasks/setup.sh
silverjam: Would like to leave it there in the hope that we can use it in the future, I think if we update `setup.sh` in the future we can propagate it's use.  With this (if used everywhere) we'd only need to update one line if the script moves, right now we'd need to update a bunch of relative paths if the script moves within the repo.

.travis.yml
silverjam: In any rate, `setup.sh` is definitely the "do bad stuff to your system" script, so installing conda there definitely felt like the right place.

tn-exafore: I agree that conda setup belongs to setup.sh. But I kind of liked the direct invocation of tox from travis. Just a personal opinion though, not necessarily the best way.

PR #1002, opened 2019-04-02
Linux pyqt5 for v. 2.3.0 [EX-468][EX-537]

tn-exafore:
Porting the pyqt5 changes from master to release 2.3.0 branch.


tn-exafore:
yep

PR #1003, opened 2019-04-02
Revert "setuptools_scm version output now includes 'v' so stop adding it (#995)"

benjaminaltieri:
Reverts #995 since release branch versions don't include the `v` by default

This reverts commit 883a1685805b06653e53209074883364ebd6b2f1.

### DIFF
--- a/piksi_tools/console/console.py
+++ b/piksi_tools/console/console.py
@@ -113,7 +113,7 @@ def get_args():
     return parser
 
 
-CONSOLE_TITLE = 'Swift Console ' + CONSOLE_VERSION
+CONSOLE_TITLE = 'Swift Console v' + CONSOLE_VERSION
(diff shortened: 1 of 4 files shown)


cbeighley:
@benjaminaltieri Looks good. The control flow is janky and complicated around this stuff but looking at https://github.com/swift-nav/piksi_tools/blob/v2.3.0-release/piksi_tools/console/update_view.py#L570 I think this should be okay.

benjaminaltieri:
@cbeighley actually I think you have pointed out that these changes are superficial and won't actually work. the underlying issue is that our new `_version.py`/`CONSOLE_VERSION` generation doesn't add a `v` for release tags, even though it does for dev/master. This PR will not solve that so I will close for now

PR #1007, opened 2019-04-03
Add tentative gui support for python 3.7 on linux [EX-541]

tn-exafore:
* also streamlines tox config a bit


requirements_gui.txt
```diff
@@ -1,7 +1,7 @@
-# fixes a memory leak. TODO: update to 4.7.3 when released
-git+git://github.com/enthought/enable@2c0a312f0b56a#egg=enable
+# TODO: move back to enthought (upstream) versions once they add python 3.7 support
+git+git://github.com/swift-nav/enable@swift-2019.04#egg=enable
```
pmiettinen: https would be better for public repos to not to force ssh key setup?

tox.ini
```diff
@@ -1,30 +1,27 @@
 [tox]
-envlist = py27,py35,py37
+envlist = {py27,py35,py37}{,-gui}
 minversion = 2.0
 
 [testenv]
 deps = -r{toxinidir}/requirements.txt
-commands = py.test -v --ignore=tests/test_console.py tests/
+       gui-py27: pyside==1.2.4
+       gui-!py27: pyqt5==5.10.0
```
pmiettinen: What's the etymology of `!`?

requirements_gui.txt
tn-exafore: We definitely don't want to require ssh key setup but I don't think this usage in requirements.txt needs it...

tox.ini
tn-exafore: that `!py27` is not a dash-limited part of the environment name

requirements_gui.txt
```diff
@@ -1,7 +1,7 @@
...
+git+git://github.com/swift-nav/chaco@swift-2019.01#egg=chaco
```
silverjam: This is a random pull of chaco master, we should probably do the same Cython update exercise on 4.7.2 before we ship this.

tox.ini
```diff
@@ -1,30 +1,27 @@
 [tox]
-envlist = py27,py35,py37
+envlist = {py27,py35,py37}{,-gui}
```
silverjam: This looks super obscure, would prefer an explicit list of everything instead so non-tox experts and quickly grok this.

tox.ini
silverjam: Again, would prefer an explicit list instead of magic `!`.

requirements_gui.txt
silverjam: I added a tag `4.7.0.dev64+swift0` so we can pin to a specific commit without a git hash and not use the branch name.

requirements_gui.txt
silverjam: I added a tag `4.7.2+swift0` which updates the cython generated modules.

silverjam:
```suggestion
git+git://github.com/swift-nav/chaco@4.7.2+swift0#egg=chaco
```

requirements_gui.txt
silverjam:
```suggestion
git+git://github.com/swift-nav/enable@4.7.0.dev64+swift0#egg=enable
```

tn-exafore:
Rebased, leaving out the changes to pyinstaller tox targets as they would complicate things when explicitly listing envs, effectively making the diff a bit smaller.

tn-exafore:
I won't backport this to v2.3.0-release unless explicitly requested by somebody.

PR #1009, opened 2019-04-04
Update Windows builds to python 3.5 [EX-539]

tn-exafore:
This is the final platform to be updated to python 3, allowing some minor cleanup in build configs.

**NOTE: if nobody objects to it, once this PR has passed reviews and been merged to master, I'll immediately backport this to `v2.3.0-release` without a separate review if the backporting doesn't require any significant adaptation.**


pmiettinen:
Windows installation package size is over 300MB now (~doubled what it used to be). I think it's partly because of numba and partly because of this change.

pmiettinen:
Current master including numba produces 226MB package for windows.

tn-exafore:
Probably pyqt5 increases the size.

denniszollo:
Looks good to me, but I think:
  - we should try and test bundled installer as soon as possible on windows7 32bit, windows7 x64, windows 8.1 x64 and windows10 x64.  I have had a lot of "side by side" errors in past due to mismatched msvc dependencies when working on the windows build and past
  - we should remove Qt5WebEngineCore.dll with the --exclude option in pyinstaller which takes up bulk of space
image

silverjam:
Size increase is a little concerning, but I'm not sure how much an issue it'll be...  if we need to address it we can open a new PR.  Merging so we can cut a new 2.3 release today.

PR #1020, opened 2019-04-12
(Experimental) Keep to older Qt scaling mechanism under Windows [ESD-1232]

tn-exafore:
Otherwise the layout changes a bit and plot legends can become
too small to read when using Windows screen scaling.

### DIFF
--- a/piksi_tools/console/windows_scaling.py
+++ b/piksi_tools/console/windows_scaling.py
@@ -19,6 +19,17 @@ import pyface.qt
 #
 # Ideally would become unnecessary one day when all components, plots included,
 # would properly work in a high dpi environment.
+
+if '--print-dpi' in sys.argv:
+    from PyQt5 import QtWidgets
+
+    app = QtWidgets.QApplication([])
+    dpi_list = [screen.logicalDotsPerInch() for screen in app.screens()]
+    max_dpi = max(dpi_list)
+
+    print(max_dpi, end="")
+    sys.exit(0)
+
 if sys.platform == "win32" and pyface.qt.qt_api.lower() in ['pyqt5', 'pyside2']:
     prog_str = b"""\
 from PyQt5 import QtWidgets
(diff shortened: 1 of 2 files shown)


silverjam:
These seems to have a pretty negative effect on machines with low DPI.  On a machine with low DPI and without this fix:

plot legend - low dpi

And, on a low DPI machine *with* this fix:

plot legend - low dpi - PR1020

High DPI machine without this fix (though, not as high as the surface):

plot legend - high dpi

Then with this fix:

plot legend - high dpi - pr1020

tn-exafore:
Tweaked the PR so that it uses the simplified scaling only on high-dpi windows machines. Let me know if works better. If it still looks bad, please add dpi info of your monitor.

piksi_tools/console/windows_scaling.py
```diff
@@ -0,0 +1,43 @@
...
+app = QtWidgets.QApplication([])
+dpi_list = [screen.logicalDotsPerInch() for screen in app.screens()]
+max_dpi = max(dpi_list)
+
+print(max_dpi, end="")
+"""
+    # initializing the application to use this Windows scaling mechanism is
+    # only possible before initializing the QApplication for the first time,
+    # so the dpi query, which also uses Qt, is done in a separate process
+    helper_proc = subprocess.Popen([sys.executable, "-u", "-"],
```
silverjam: I think once piksi_tools is reduced to a single executable by PyInstaller `sys.executable` points to `console.exe`-- I'm currently seeing that this logic "fork bombs" the computer by repeatedly launching `console.exe`.  Which results in error dialogs like this:

low-dpi-scaling-fix-error

low-dpi-scaling-fix-error2

tn-exafore:
Damn, I didn't test the pyinstaller bundle. A bit weird that it works so differently to the unbundled python. Need to fix that, obviously.

tn-exafore:
@silverjam I hacked the pyinstaller problem with another hack (broke python 2.7 but that's easy to fix next week if this helps otherwise). If you have time, have a try.

silverjam:
@tn-exafore I'm still seeing some fuzziness on low dpi systems with the most recent PR.  This is with this PR:

console-2 3 8-lowdpi-fuzz

Versus without this PR:

console-2 3 8-lowdpi

tn-exafore:
@silverjam I'm starting to feel that this isn't a working approach for all possibilities. Also the code is getting so ugly that I wouldn't be particularly eager to merge it. Created another PR, #1036, that does the thing more manually.

PR #1024, opened 2019-04-15
Update settings docs

silverjam:
Add docs to squelch these warnings:
```
No settings documentation entry for name connectivity_check_addresses and group is system
No settings documentation entry for name gpgst_msg_rate and group is nmea
```

### DIFF
--- a/piksi_tools/console/settings.yaml
+++ b/piksi_tools/console/settings.yaml
@@ -1501,6 +1501,17 @@
   Description: Number of Solution Periods between GSA NMEA messages being sent.
   Notes: This setting represents the integer number of solution periods between each transmission of the NMEA message.
 
+- group: nmea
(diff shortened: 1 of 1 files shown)


switanis:
People may get confused about quotes in description:
'8.8.8.8,1.1.1.1'
I'd remove them.

Suggestion:
A value of 0 will disable the connectivity check and Link LED will not show Internet access status.


For NMEA I'd remove Talker ID, i.e. "gp". Just gga, gst, ... is enough. But that'd be a separate issue for all entries I guess.

I'd also say zero will disable message output.

silverjam:
@switanis Please log an ESD issue for removing "gp", we'll need to make code changes to the firmware to accomplish this.  Other feedback incorporate in https://github.com/swift-nav/piksi_tools/pull/1029.

PR #1028, opened 2019-04-17
Tweak the "Refresh settings from device" button [ESD-1268]

tn-exafore:
The previously separate button text moved into a button label to be
similar with the other buttons.

There is a bug in enable's SVGButton that horizontal layout doesn't
work (probably the reason why this button was originally configured
separately for button and text). I made a separate PR to fix that at https://github.com/enthought/enable/pull/338 and they already approved & merged it. I updated the swift fork's master to reflect this but didn't yet update requirements_gui.txt. Wondering whether that dependency should be updated just for master or for both master and v.2.3.0-release...

### DIFF
--- a/piksi_tools/console/gui_utils.py
+++ b/piksi_tools/console/gui_utils.py
@@ -14,6 +14,8 @@ from pyface.api import GUI
 
 from traits.api import Bool, HasTraits, List
 from traitsui.api import HGroup, VGroup, Item, TextEditor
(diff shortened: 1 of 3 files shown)


tn-exafore:
@silverjam: Updated reqs and also modified the "Show advanced settings" checkbox functionally the same way (although the implementation is different).

tn-exafore:
I'll backport this to release branch next week if no problems found.

PR #1029, opened 2019-04-17
Fix settings docs YAML error, incorporate feedback [ESD-1283]

silverjam:
Fix YAML parse snafu, incorporate feedback from @switanis.


silverjam:
GUI smoke test passed, merging...

PR #1033, opened 2019-04-23
Disable drag & drop in tables [ESD-1282]

tn-exafore:
* also disable editing in some tables where it was previously implicitly allowed but makes no sense

### DIFF
--- a/piksi_tools/console/gui_utils.py
+++ b/piksi_tools/console/gui_utils.py
@@ -16,6 +16,7 @@ from traits.api import Bool, HasTraits, List
 from traitsui.api import HGroup, VGroup, Item, TextEditor
 from traitsui.basic_editor_factory import BasicEditorFactory
 import traitsui.qt4.boolean_editor
+from traitsui.tabular_adapter import TabularAdapter
 
 from piksi_tools.console.utils import SUPPORTED_CODES, GUI_CODES, code_to_str
 
@@ -146,3 +147,12 @@ class PiksiBooleanEditor(BasicEditorFactory):
        by making the UI item's label part of the toggle button (editor).
        This enables toggling it by clicking on the text, too, as in normal Qt."""
     klass = _PiksiBooleanEditor
+
+
+class ReadOnlyTabularAdapter(TabularAdapter):
+    '''A tabular editor that is by default uneditable and where drag
+       (of drag & drop) is disabled'''
+    can_edit = Bool(False)
+
+    def get_drag(self, object, trait, row):
+        return None  # disables drag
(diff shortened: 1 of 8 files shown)


piksi_tools/console/system_monitor_view.py
```diff
@@ -32,10 +32,18 @@ def ip_bytes_to_string(ip_bytes):
...
     columns = [('Thread Name', 0), ('CPU %', 1), ('Stack Free', 2)]
+    can_edit = Bool(False)
+
+    def get_drag(self, object, trait, row):
+        return None  # disables drag
 
 
 class SimpleCSACAdapter(TabularAdapter):
     columns = [('Metric Name', 0), ('Value', 1)]
+    can_edit = Bool(False)
```
benjaminaltieri: this looks like shotgun surgery, can class level variables be inherited? maybe a superclass that all of these SimpleAdapters can inherit would reduce duplicated logic here.

tn-exafore: Agreed, it was a bit rushed. Refactored.

tn-exafore:
I'll "automatically" backport this to release-v2.3.0 when this PR is approved if nobody objects to it.

piksi_tools/console/output_list.py
```diff
@@ -20,7 +20,7 @@
 from traits.api import (Bool, Enum, Float, Font, HasTraits, Int, List,
                         Property, Str, Trait)
 from traitsui.api import TabularEditor, UItem, View
-from traitsui.tabular_adapter import TabularAdapter
+from .gui_utils import PiksiTabularAdapter
```
benjaminaltieri: full path import

piksi_tools/console/solution_view.py
```diff
@@ -34,7 +34,7 @@
                         Str, Enum)
 from traitsui.api import (HGroup, HSplit, Item, TabularEditor, TextEditor,
                           VGroup, View)
-from traitsui.tabular_adapter import TabularAdapter
+from .gui_utils import PiksiTabularAdapter
```
benjaminaltieri: Same here

piksi_tools/console/gui_utils.py
```diff
@@ -130,3 +131,10 @@ def get_filter_group():
                         visible_when="{} in received_codes".format(code)))
             hgroup.content.append(vgroup)
         return hgroup
+
+
+class PiksiTabularAdapter(TabularAdapter):
```
benjaminaltieri: this looks great, piksi is a bit of a misnomer though, maybe `NoEditTabularAdapter`? Or something version that denotes what is being inherited

tn-exafore: There is some chance of further style etc. updates in the future. Anyway, renamed to `ReadOnlyTabularAdapter`, it is a descriptive name at the moment.

piksi_tools/console/output_list.py
tn-exafore: fixed

PR #1034, opened 2019-04-24
Use monotonic time to prevent plotting artifacts [ESD-1269]

tn-exafore:
Decrementing system time causes plotting artifacts in tracking view. This can be fixed by switching to a monotonic clock  that never goes backwards. This is a wrap-up of https://github.com/swift-nav/piksi_tools/commit/3afede840b6e87fe33fd7867a2346200ca997c5c.

### DIFF
--- a/piksi_tools/console/baseline_view.py
+++ b/piksi_tools/console/baseline_view.py
@@ -11,6 +11,7 @@
 import datetime
 import os
 import time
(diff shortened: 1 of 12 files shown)


tn-exafore:
I'll "automatically" backport this to release-v2.3.0 when this PR is approved if nobody objects to it.

PR #1036, opened 2019-04-25
(WIP) Add alternatively scaled windows shortcuts [ESD-1232]

tn-exafore:


denniszollo:
Regarding the multiple shortcuts, I think it would be better if the second shortcut only shows up in program files, and we just keep one shortcut on the Desktop.  We can put in the installation instructions and documentation how to "replace" the shortcut on the desktop with the preferred variant if desired.  I think it might be strange for Windows users to get two different shortcuts on the desktop after installation.

tn-exafore:
> 
> 
> Regarding the multiple shortcuts, I think it would be better if the second shortcut only shows up in program files, and we just keep one shortcut on the Desktop. We can put in the installation instructions and documentation how to "replace" the shortcut on the desktop with the preferred variant if desired. I think it might be strange for Windows users to get two different shortcuts on the desktop after installation.

Ok, I'll update this and the other variants that way today.

PR #1037, opened 2019-04-26
Remove a broken optimization to fix computed doppler [ESD-1276]

tn-exafore:
Otherwise local computed doppler is missing when setting (soln freq) / (output every n obs) is high. The optimization was also otherwise broken in that it didn't update TOW as advertised.

Also, UI updates are nowadays separately scheduled so this doesn't directly affect UI update frequency.

IMO it is best to remove this broken optimization. If that is not acceptable, it's also possible to keep & fix  it, but it would be more complex (keeping track of collecting computed doppler), and IMO this was/is a slightly dangerous optimization, could cause also other subtle bugs.

I'll backport this to v2.3.0-release upon approval of this PR.

### DIFF
--- a/piksi_tools/console/observation_view.py
+++ b/piksi_tools/console/observation_view.py
@@ -170,10 +170,7 @@ class ObservationView(CodeFiltered):
             return
         else:
             self.prev_obs_count = count
-        # Don't bother updating anything except the TOW faster than 2Hz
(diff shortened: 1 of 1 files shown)


denniszollo:
Looks good to me. These things were all kludges to reduce CPU utilization.  I think in the end it wasn't the drawing of the observations that used to be CPU intensive, it was the unpacking of the SBP messages.

PR #1046, opened 2019-05-10
Add a new 'Sky Plot' tab

ebethon:
It displays the azimuth and elevation of tracked satellites

Co-authored-by: David Feng <davidfengabc@users.noreply.github.com>

### DIFF
--- a/piksi_tools/console/console.py
+++ b/piksi_tools/console/console.py
@@ -48,6 +48,7 @@ from piksi_tools.console.output_list import (
 from piksi_tools.console.sbp_relay_view import SbpRelayView
 from piksi_tools.console.settings_view import SettingsView
 from piksi_tools.console.solution_view import SolutionView
+from piksi_tools.console.skyplot_view import SkyplotView
 from piksi_tools.console.spectrum_analyzer_view import SpectrumAnalyzerView
 from piksi_tools.console.system_monitor_view import SystemMonitorView
 from piksi_tools.console.tracking_view import TrackingView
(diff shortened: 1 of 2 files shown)


ljbade:
Nice work, code looks fine

silverjam:
Looks a bit squished with the default resolution and the NESW labels are cut off.  Don't suppose there's any way to move the labels in a bit?

Screen Shot 2019-05-14 at 7 29 27 PM

nsirola:
Sweet. Note that SBP message 151 needs to be enabled for this to show.

denniszollo:
Code looks good, but please verify that the minimum width of the console has not changed with the addition of this tab.  Ideally, you can resize the console down so it can take up only half of a typical laptop screen.

ebethon:
I just pushed some cosmetic changes if you want to check that out

silverjam:
@ebethon Looks better, do you think it'd be easy to add a note about what messages are needed to enable the plot?  On the "spectrum analyzer" plot we have a note that says `Enable with settings in "System Monitor" group` -- if someone has persisted their enabled message list they won't see anything from the Sky Plot until they add message 151 (and it doesn't look like message 151 is enabled by default).  So, I think there's 2 potential concerns here: (1) we should probably add some help text to the plot for which messages are needed and (2) should we enable message 151 by default in Piksi?

ebethon:
I think we should add the message 151 in the default messages, it is pretty lightweight (8 + 4N bytes, where N is the number of tracked satellites => 96 bytes max). In that case, we don't need a helper.

If we don't want to enable the message 151 by default, we should move that tab under advanced and add a helper as you recommend.

silverjam:
I will advocate for moving the plot to advanced then, unless there are other use cases for message 151.  We are already at limit of the UART data rate such that we occasionally go over it with the current default message configuration (especially when lots of satellites are in view).  We need to change our default UART rate, or our default message mask to give us more head room... because of this, I think it makes more sense to move this to advanced.  Thoughts?

ebethon:
@ljbade @silverjam @denniszollo I moved the 'Sky Plot' tab under the 'Advanced Tab' with a text hint indicating the user needs to add the message 151.

silverjam:
@ebethon Let's at least merge to master now? Still have a few days to debate 2.3...

PR #1051, opened 2019-05-17
Add ethernet:interface_mode setting to yaml [ESD-1320]

yizow:
Add description for new ethernet:interface_mode setting

### DIFF
--- a/piksi_tools/console/settings.yaml
+++ b/piksi_tools/console/settings.yaml
@@ -1704,16 +1704,16 @@
 - group: ethernet
   name: mode
   expert: false
-  type: boolean
-  enumerated possible values: True,False
+  type: enum
+  enumerated possible values: Config,Active
   units: N/A
-  default value: 'False'
+  default value: 'Active'
   readonly: false
-  Description: Ethernet configuration locked toggle.
+  Description: Ethernet configuration mode.
   Notes: |
-    "Config" IP configuration can be changed freely, but no change is made on the device.
+    "Config" IP configuration can be changed freely, but no change is made on the device. Returning to 'active' mode will refresh ethernet connection with current values.
 
-    "Active" The current IP configuration is sent to the device and updated. Afterward, no IP settings can be changed until returned to 'config' mode
+    "Active" The current IP configuration is sent to the device and updated. Afterward, no IP settings can be changed until returned to 'config' mode.
 
 - group: ethernet
   name: ip_config_mode
--- a/piksi_tools/console/settings.yaml
+++ b/piksi_tools/console/settings.yaml
@@ -1702,7 +1702,7 @@
   Notes:
 
 - group: ethernet
-  name: locked
+  name: mode
   expert: false
   type: boolean
   enumerated possible values: True,False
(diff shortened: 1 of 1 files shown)


piksi_tools/console/settings.yaml
```diff
@@ -1701,6 +1701,20 @@
   Description: Enable the almanac-based acquisition.
   Notes:
 
+- group: ethernet
+  name: mode
+  expert: false
+  type: boolean
+  enumerated possible values: True,False
```
benjaminaltieri: type should be enum, values `active` or `config`

piksi_tools/console/settings.yaml
```diff
@@ -1701,6 +1701,20 @@
...
+  units: N/A
+  default value: 'False'
```
benjaminaltieri: needs update

piksi_tools/console/settings.yaml
```diff
@@ -1701,6 +1701,20 @@
...
+  readonly: false
+  Description: Ethernet configuration locked toggle.
```
benjaminaltieri: "Ethernet configuration mode:"

piksi_tools/console/settings.yaml
```diff
@@ -1701,6 +1701,20 @@
...
+  Notes: |
+    "Config" IP configuration can be changed freely, but no change is made on the device.
```
benjaminaltieri: Append "Changing setting to Active will refresh ethernet connection with current values."

denniszollo: I can't determine how this setting works from the description.  If in "active" mode, does the device behave as it used to on prior releases?

benjaminaltieri: Underlying behavior is that we want 'active' to denote currently active settings and delay application of settings changes to individual settings until it is desired to do so. The solution is to change to 'config' mode which allows you to change the setting without them taking effect. Switching back to 'active' will apply the settings and re-up the interface in one go

piksi_tools/console/settings.yaml
```diff
@@ -1701,6 +1701,20 @@
...
+  name: interface_mode
+  expert: false
+  type: enum
+  enumerated possible values: Config,Active
+  units: N/A
+  default value: 'Active'
+  readonly: false
+  Description: Ethernet configuration mode.
+  Notes: |
+    "Config" IP configuration can be changed freely, but no change is made on the device. Returning to 'active' mode will refresh ethernet connection with current values.
```
silverjam:
```suggestion
    "Config" IP configuration can be changed freely, but no change is made on the device. Returning to 'Active' mode will refresh ethernet connection with current values.
```

piksi_tools/console/settings.yaml
```diff
@@ -1701,6 +1701,20 @@
...
+
+    "Active" The current IP configuration is sent to the device and updated. Afterward, no IP settings can be changed until returned to 'config' mode.
```
silverjam:
```suggestion
    "Active" The current IP configuration is sent to the device and updated. Afterward, no IP settings can be changed until returned to 'Config' mode.
```

PR #1052, opened 2019-05-20
Add NetworkCheck class [ESD-1310]

pmiettinen:
https://swift-nav.atlassian.net/browse/ESD-1310

Bonus option; Add an icon showing the network status similar to the device connection icon.

### DIFF
--- a/piksi_tools/console/update_view.py
+++ b/piksi_tools/console/update_view.py
@@ -64,7 +64,7 @@ class NetworkCheck():
     def __init__(self, log, cb):
         self._cb = cb
         self._log = log
-        self._prev_state = True
+        self._state = True
         self._thd_trigger = Event()
         self._check_done = Event()
         self._worker = Thread(target=self._check_thd)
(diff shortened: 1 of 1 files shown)


denniszollo:
Can you explain motivation for this?  Is the goal to have more relevant error messages when downloading things from internet on console?

benjaminaltieri:
@dzollo the objective is to avoid the behavior reported by @switanis in https://swift-nav.atlassian.net/browse/ESD-1310
As stated in the ticket:
```
Console v2.3.8 shows error when PC is not on Internet (on Update tab):

Error: Failed to download latest file index from Swift Navigation's website. Please visit our website to check that you're running the latest Piksi firmware and Piksi console.

Error: No website index to use to compare versions with local firmware

Seen on Win7 PC.

Such error should not be displayed. This is not an error.
```

denniszollo:
Thanks Ben ; I should have seen the ticket in the title!

piksi_tools/console/update_view.py
```diff
@@ -54,6 +55,46 @@ def parse_version(version):
     return gitversion_parse(version)
 
 
+class NetworkCheck():
+
+    CONNECTION_TIMEOUT = 1.0
+    THREAD_INTERVAL = 5.0
```
silverjam: Can we bump this to 20 or 30 seconds?  A 5 second interval will make the console consume about 2mb a day over a 24hr period of Internet traffic (when it previously only consumed Internet traffic when directed, or when checking the firmware version, afaik).

@denniszollo @switanis Do you think it would be a concern for customers if the console started to ping www.google.com regularly?

piksi_tools/console/update_view.py
```diff
@@ -54,6 +55,46 @@ def parse_version(version):
...
+
+    def _check_thd(self):
+        while True:
+            try:
+                socket.create_connection(NetworkCheck.SERVER, NetworkCheck.CONNECTION_TIMEOUT).close()
+                if not self._prev_state:
+                    self._log("Network is reachable.")
+                self._prev_state = True
+                self._cb(True)
+            except Exception as ex:
```
silverjam: We don't need to capture `ex` if we're not using it, and we should probably make the `try/catch` just wrap the call that we expect to fail (`socket.create_connection`), since we're using the very general `Exception`.  Unless there was some other rational for this?

piksi_tools/console/update_view.py
```diff
@@ -54,6 +55,46 @@ def parse_version(version):
...
+                if self._prev_state:
+                    self._log("Network is unreachable, please check your connection.")
```
silverjam: This seems like it will be very noisy in situations where there's no Internet connectivity, like a drive test?  Same with the inverse message above.

piksi_tools/console/update_view.py
pmiettinen: lint didn't approve the bare `except:`.

piksi_tools/console/update_view.py
pmiettinen: Log is printed only when the state changes.

benjaminaltieri:
Need to decide whether or not to backport to 2.3

PR #1055, opened 2019-05-20
Add console downloader functionality to console

denniszollo:
This pull requests adds the following features to the console.  Both features represent more or less completed development that I wanted to tidy up for possible merging.

- If the console is outdated, it prompts the user to download with the following popup.
image

- It adds a new "Download Latest Console" button on update view
image

If user presses ok in the first popup or the new button, a file chooser dialog opens up to choose a location to download.  Then the console installer for the platform is downloaded.  During download, periods print across the screen to indicate that something is happened a status message appears in Firmware upgrades status window per below when finished.

image

### DIFF
--- a/piksi_tools/console/update_view.py
+++ b/piksi_tools/console/update_view.py
@@ -97,6 +97,7 @@ class FirmwareFileDialog(HasTraits):
             label='Choose Firmware File',
             action='open',
             default_directory=self.default_dir,
+            default_path=self.default_dir,
             wildcard=self.file_wildcard)
         dialog.open()
         if dialog.return_code == OK:
@@ -183,6 +184,9 @@ class UpdateView(HasTraits):
     download_directory_label = String('Firmware Download Directory:')
 
     update_stm_firmware = Button(label='Update Firmware')
+    download_console = Button(label='Download Latest Console')
+    console_directory = Unicode()
+    choose_console_dir = Button(label='...', padding=-1)
 
     updating = Bool(False)
     update_stm_en = Bool(False)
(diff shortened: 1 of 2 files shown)


denniszollo:
I guess at this late date I should just rebase and merge into master.

piksi_tools/console/update_view.py
```diff
@@ -497,6 +538,58 @@ def _download_firmware(self):
...
+                if e.errno == errno.EACCES or e.errno == errno.EPERM:
+                    self._write("IOError: unable to write to path %s. "
+                                "Verify that the path is writable." %
+                                path)
+                else:
+                    raise (e)
+            except KeyError:
+                self._write("Error downloading Console: URL not present in index")
+            except URLError:
+                self.nap_fw.clear("Error downloading console")
```
silverjam: Why are we touching `self.nap_fw`?

piksi_tools/console/update_view.py
```diff
@@ -497,6 +538,58 @@ def _download_firmware(self):
...
+        if self.update_dl is None or self.update_dl.index is None:
+            self._write("Error: Can't download firmware files")
+            return
+
+        status = 'Downloading Latest Console to {}'.format(path)
+        self._write(status)
+
+        # Get console files from Swift Nav's website for correct platform
+        if 'console' in self.update_dl.index[self.piksi_hw_rev]:
+            try:
```
silverjam: The KeyError and AttributeError blocks encompass very broad errors, the specific situations where these exceptions render into specific error cases should have try/catch blocks around them, and we should throw a subclass of RuntimeError with an appropriate message.

piksi_tools/console/update_view.py
```diff
@@ -565,30 +658,64 @@ def _compare_versions(self):
...
+            if self.fw_outdated:
+                fw_update_prompt = \
+                    prompt.CallbackPrompt(
+                        title='Firmware Update',
+                        actions=[prompt.close_button]
+                    )
+
+                if 'fw' in self.update_dl.index[self.piksi_hw_rev]:
+                    fw_update_prompt.text = \
+                        "New Piksi firmware available.\n\n" + \
```
silverjam: Move to top-level multi-line string constant and use `str.format` instead of `%`?  See https://pyformat.info/ for more details.

PR #1057, opened 2019-05-24
UpdateScheduler thread safety [ESD-1239]

pmiettinen:


benjaminaltieri:
just saw your comment @silverjam - feel free to request changes if you think this isn't the right approach. but considering @nsirola's comment in https://swift-nav.atlassian.net/browse/ESD-1239 it seems worth merging until a more efficient solution is identified

PR #1061, opened 2019-06-10
fileio: fix iter over changing dict, bug in time diff util

denniszollo:
Proposed fix for https://swift-nav.atlassian.net/browse/ESD-1365

Also fixes bug in handling difference between the milliseconds field when subtracting times.  (this caused negative file transfer rates reported on small text files).

### DIFF
--- a/piksi_tools/utils.py
+++ b/piksi_tools/utils.py
@@ -96,7 +96,7 @@ class Time(object):
             yield next_time
 
     def to_float(self):
-        return self._seconds + (self._millis / 1000)
+        return self._seconds + (self._millis / 1000.0)
 
     def __hash__(self):
         return hash((self._seconds, self._millis))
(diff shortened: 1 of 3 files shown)


piksi_tools/utils.py
```diff
@@ -109,7 +109,7 @@ def __add__(a, b):
         return Time(seconds=(new_time // 1000), millis=(new_time % 1000))
 
     def __sub__(a, b):
-        new_time = (1000 * a._seconds) - (1000 * b._seconds) - a._millis - b._millis
+        new_time = (1000 * a._seconds) - (1000 * b._seconds) + (a._millis - b._millis)
```
silverjam: Please add a unit test for this:

```diff
diff --git a/tests/test_utils.py b/tests/test_utils.py
index f189d422..5f33f004 100644
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -22,6 +22,12 @@ def test_time_add():
     assert (t4 + t5) == Time(3, 0)


+def test_time_sub():
+    t4 = Time(1, 400)
+    t5 = Time(1, 300)
+    assert (t4 - t5) == Time(0, 100)
+
+
```

PR #1091, opened 2019-09-11
Update Dockerfile from trusty (14.04) to bionic (18.04)

dgburr:
The previous version no longer builds due to new dependencies in requirements.txt.


silverjam:
closed since we merged #1077

PR #1098, opened 2019-11-08
Ubuntu 18.04 support master

denniszollo:


denniszollo:
This is no good; somehow the auto-cherry-pick are missing things that were force pushed later. I am manually cherry-picking then will force push.

PR #1113, opened 2019-12-28
Add user-configurable message filters to sbp_log_zipper

dgburr:
Add command line arguments --base-filter and --rover-filter to allow the user to specify custom message filters.


silverjam:
this tool has been deprecated/removed in #1139 -- we should prefer the new interleaver here: https://github.com/swift-nav/interleaver

PR #1123, opened 2020-01-24
[DML-23] Solution tab zooming into singularity 

pmiettinen:
When PVT is pinned, add epsilon to prevent zero range in plotting axes.

image

image


piksi_tools/console/gui_utils.py
```diff
@@ -8,6 +8,8 @@
 # EITHER EXPRESSED OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE IMPLIED
 # WARRANTIES OF MERCHANTABILITY AND/OR FITNESS FOR A PARTICULAR PURPOSE.
 
+import sys
```
nsirola: unused import?

pmiettinen: Leftover from trying different approaches, removing..

PR #1127, opened 2020-02-05
2.4.0 release rearrange tabs, add velocity

denniszollo:
1. Moves skyplot into tracking view as a subtab:
image
2. Add velocity graph as a subtab:
image

### DIFF
--- a/piksi_tools/console/console.py
+++ b/piksi_tools/console/console.py
@@ -227,7 +227,10 @@ class SwiftConsole(HasTraits):
     view = View(
         VSplit(
             Tabbed(
-                Item('tracking_view', style='custom', label='Tracking'),
+                Tabbed(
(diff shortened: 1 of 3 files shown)


piksi_tools/console/console.py
```diff
@@ -227,7 +227,10 @@ class SwiftConsole(HasTraits):
     view = View(
         VSplit(
             Tabbed(
-                Item('tracking_view', style='custom', label='Tracking'),
+                Tabbed(
+                    Item('tracking_view', style='custom', label='CN0 Graph', show_label=False),
```
switanis: I'd call is SV Signals or just Signals

denniszollo: you're thinking "Signals" and "Skyplot" for the labels?

PR #1131, opened 2020-02-10
add data rate display to console and serial_link

denniszollo:
Adds data rate display to swift console in Kilobytes per second:
 
requires https://github.com/swift-nav/libsbp/pull/775

Looks like this if disconnected completely:
image
This is connected to something but there is no SBP_MSG_HEARTBEAT
image

and this if connected and has heartbeat:
image

Averaging period is 1.2 seconds to avoid ping-ponging on either side of 1Hz GNSS epochs and misleading people.

### DIFF
--- a/piksi_tools/serial_link.py
+++ b/piksi_tools/serial_link.py
@@ -280,7 +280,7 @@ def swriter(link):
     return scallback
 
 
-def run(args, link):
+def run(args, link, stop_function=None):
     """Spin loop for reading from the serial link.
 
     Parameters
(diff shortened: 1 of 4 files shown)


piksi_tools/serial_link.py
```diff
@@ -275,7 +280,7 @@ def scallback(sbp_msg, **metadata):
     return scallback
 
 
-def run(args, link):
+def run(args, link, stop_function=None):
```
silverjam, line 283: If this is just `stop_function=lambda: None` you won’t need any of the `if stop_function is not None` checks.

silverjam:
@dzollo merge?

PR #1136, opened 2020-02-17
[DML-26] Limit float precision in settings view

pmiettinen:
## Motivation
Discussion here https://github.com/swift-nav/piksi_buildroot_private/pull/270#issuecomment-586460881

Basically it's confusing for the user to see "too precise" values.

This has probably been considered before but I am not familiar with earlier discussions? Maybe this is just a bad idea in general to have different value presented than the actual value is? To be discussed.

## Root of the problem

Floating point precision. Floats (and doubles) have limited precision and can't precisely present values like `0.1`. Instead the closest value will be selected.

## Implementation

Without PR:
image

With PR:
image

## TODO/TBD
- [ ] What should happen if user enters more digits than defined?


piksi_tools/console/settings_view.py
```diff
@@ -277,6 +293,22 @@ class SimpleAdapter(ReadOnlyTabularAdapter):
...
+        if 0 == column:
+            return super(SimpleAdapter, self).get_text(object, name, row, column)
+
+        settings_list = getattr(object, 'settings_list')
+        setting = settings_list[row]
+
+        digits = getattr(setting, 'digits', None)
+
+        if digits:
+            return f'%.{digits}f' % float(setting.value)
```
pmiettinen: Doesn't work on all py versions..

silverjam:
I suppose the alternative would be to change the units of the settings to tenths of a second, so we don't have to do this.

pmiettinen:
Currently breaks the EnumSetting behavior, needs some elaboration.

pmiettinen:
Suggestion; Show the actual value on device in addition to the truncated value:

image

piksi_tools/console/settings.yaml
```diff
@@ -294,6 +294,7 @@
   name: phase_sigma
   expert: false
   type: double
+  digits: '3'
```
pmiettinen: Demo/testing change, can be reverted.

silverjam:
@pmiettinen merge at your leisure, then port to 2.4?

PR #1137, opened 2020-02-19
Expand sbp json payload

pmiettinen:
## Motivation

I needed to expand SBP JSON log payloads here https://swift-nav.atlassian.net/browse/DML-20

I think I have encountered the situation before, but could not recollect how I approached the problem then. Also could not find support for this in the existing tools. Two separate solutions are now offered. Not necessary to include both.

## Initial solution
New script utilizing existing SBP iterators, see `sbpjson_expand.py`.

## Improved(?) solution
Extend `serial_link.py` to accept JSON log iterator as a source. Needs changes in `libsbp`.

## Requires
https://github.com/swift-nav/libsbp/pull/776


denniszollo:
Seems like we need to do a release or libsbp to merge this.

pmiettinen:
> looks reasonable; It might also be nice to add same argument to the console one-day so it can playback that which it logs.

Argument actually is consequently added to the console also as it reuses the serial_link argument parser. What is missing is sort of playback file driver that feeds the messages with appropriate delays with regards to timestamps.

pmiettinen:
Implementing some naive level playback driver might actually not require too much of an effort.

silverjam:
swift-nav/libsbp#779 is released in sbp version 2.7.4

silverjam:
Seems this was already merged somewhere else?  Closing.

silverjam:
Screenshot from 2020-04-10 14-50-25

PR #1138, opened 2020-02-28
Playback support in serial_link

pmiettinen:
Requires https://github.com/swift-nav/libsbp/pull/782

Based on https://github.com/swift-nav/piksi_tools/pull/1137


silverjam:
https://github.com/swift-nav/libsbp/pull/779 is released

silverjam:
Let's port to 2.4-- otherwise no one is going to know it's there or be able to use it, and it looks like a low risk thing to backport

PR #1143, opened 2020-03-12
Failed start-up on Windows [PMF-85] [INFRA-128]

silverjam:
See https://swift-nav.atlassian.net/browse/PMF-85

The bundled executable fails to start on Windows.  Upgrade to PyInstaller 3.6 and pin setuptools to prevent this from happening in the future.


silverjam:
@switanis and @nsirola confirmed that this works

switanis:
LGTM. Tested in Win7 and macOS 10.14.6.

PR #1145, opened 2020-03-13
Try PyQt 5.12.3

silverjam:


silverjam:
worked on Linux, but not on Windows, could not test on OS X

PR #1152, opened 2020-03-27
[PMF-95] Prevent write-back

pmiettinen:
https://swift-nav.atlassian.net/browse/PMF-95

Bug description:
1. Run console and Piksi with `pfwp_build_date` `a.b.c`
2. Make sure console has executed `_read_all_thread`
3. Update Piksi build with  `pfwp_build_date` `x.y.z` and reboot Piksi and do not restart console
4. Piksi reboots, console detects it and executes `_read_all_thread`
5. New `pfwp_build_date` value is assigned in the console
6. Since `pfwp_build_date` value has changed, console sends `MSG_SETTINGS_WRITE`
7. Firmware settings client receives the write request and detects that it tries to update read-only setting
8. Warning is issued and request rejected

This PR omits steps 6-8.

### DIFF
--- a/piksi_tools/console/settings_view.py
+++ b/piksi_tools/console/settings_view.py
@@ -178,7 +178,7 @@ class Setting(SettingBase):
         if self.readonly:
             return
 
-        self.update_value(value, False)
+        self.update_value(old, False)
 
         invalid_setting_prompt = prompt.CallbackPrompt(
             title="Settings Write Error: {}.{}".format(section, name),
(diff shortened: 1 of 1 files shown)


piksi_tools/console/settings_view.py
```diff
@@ -260,6 +258,11 @@ def _value_changed(self, name, old, new):
 
         self.settings.workqueue.put(self._write_value, old, new)
 
+    def update_value(self, new, send_write_req):
```
silverjam: When would be ever pass `True` for `send_write_req`?

pmiettinen:
@silverjam I added more detailed PR description.

silverjam:
Should this be in `v2.4.0-release`?

pmiettinen:
It's not mission critical but would enhance the release quality. I'll make a PR.

PR #1155, opened 2020-04-22
Do not use baseline view last_mode unless in DGNSS mode

pmiettinen:
Mitigate status bar fix type field defect found in https://swift-nav.atlassian.net/browse/PMF-36

Note that this doesn't change the baseline view aspect, it's still showing the last received data.

Not super pleased with this change but I think this is one of the least intrusive hackerinos to mitigate the status bar bug. Probably would be a good idea to separate the fix types between solution and baseline if they can differ since it's hard/impossible to do a logical implementation if two concepts are forced into one space.

### DIFF
--- a/piksi_tools/console/console.py
+++ b/piksi_tools/console/console.py
@@ -479,7 +479,7 @@ class SwiftConsole(HasTraits):
                 llh_display_mode += "+INS"
 
         # determine the latest baseline solution mode
-        if self.baseline_view:
+        if self.baseline_view and self.settings_view and self.settings_view.dgnss_enabled():
             baseline_solution_mode = self.baseline_view.last_mode
             baseline_display_mode = mode_dict.get(baseline_solution_mode, EMPTY_STR)
             if baseline_solution_mode > 0 and self.baseline_view.last_soln:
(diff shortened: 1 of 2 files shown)


pmiettinen:
Adding another UI element is the first thing that comes to mind but I am not familiar enough with the differences between the fix types that I can confidently start implementing and expecting it to be coherent enough.

@denniszollo maybe you could explain the details of the business logic here and what is expected to happen regarding the fix types.

denniszollo:
To me the idea is to show; is RTK happening? Has it converged? And to show it in the case of moving baseline AND pseudo absolute position.

This fix seems fine because apparently the piksi stops sending out Baseline messages in any way when that settings change is made, but I feel like the status bar timeout logic should have caught this condition rather than having to check the setting explicitly.

The concept I designed seems to confuse a lot of people who are doing moving baseline (either accidentally or on purpose) or aren't getting the position from their base station (either accidentally on purpose).  I don't think a separate fix mode in the status bar warrants the real estate really since few people use moving baseline and that is when it would apply most directly.

silverjam:
Backport? Or change the target branch to 2.4?

silverjam:
Please merge, Dennis indicated that this fix is acceptable

PR #1161, opened 2020-06-08
Drive tracking view from OBS message if no tracking or measurement st…

denniszollo:
This PR drives the tracking tab from incoming MSG_OBS if no MSG_TRACKING_STATE or MSG_MEASUREMENT_STATE have yet been received.  Tracking tab should behave identically in either case.  This make the console compatible with systems that don't produce separate tracking information (i.e non piksi Multi receivers in use).

/cc @switanis @cbeighley @silverjam

### DIFF
--- a/piksi_tools/console/console.py
+++ b/piksi_tools/console/console.py
@@ -670,7 +670,8 @@ class SwiftConsole(HasTraits):
                 self.link,
                 name='Local',
                 relay=False,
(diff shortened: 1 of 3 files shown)


denniszollo:
Tested by Stefan; appears to work.

denniszollo:
We have this auto PR process for master, so once I merge, it will auto create a PR.

PR #1175, opened 2020-07-18
Dzollo/status bar v2.4

denniszollo:
small change requests from Stefan and also our attempt to try to fix the oft-confused issue where the BASELINE had RTK, but the POS is still in SPP and users are confused.  The combined mode was my design originally and enough people (customers, Exafore, new employees, etc) have complained that I have seen the light.

Used to look like this:

image
image

Now looks like this:
image (25)
image (26)

Additional changes:
I also tried to make sure that if a message stops (like pos_llh or basline_ned), that the status bar doesn't show stale data.  
I tried to make the status bar work even without MSG_HEARTBEATS

I think this one needs some testing especially with Duro Inertial / Step / hosted Starling, so will ask @switanis  to give a whirl before any merging.


switanis:
Works well for me.

PR #1210, opened 2020-11-12
Fix SR-233 by documenting heading forward setting

denniszollo:
This pr documents the heading forward setting and forward ports important things from the v2.4.0 release branch that weren't on the 3.0.0 release branch due to me not being careful with our process.


silverjam (approved):

silverjam:
@denniszollo this PR should target the v3.0.0-release branch, right?

denniszollo:
I just checked piksi_apps and see we don't forward BASELINE_ECEF: https://github.com/swift-nav/piksi_apps/blob/master/app/heading_forwarder/heading_forwarder.c#L78

Will remove from the doc string

PR #1226, opened 2021-02-26
Display outlier information from MSG_INS_UPDATES in Swift console [INFRA-383]

notoriaga:


silverjam:
@notoriaga got any test data I can use with str2str to see the lights twinkle?

notoriaga:
> @notoriaga got any test data I can use with str2str to see the lights twinkle?

`drive-test-logs-2/2021-02/2021-02-11-WheelODORoute/PGM2_CAN_v1_2_34_wheelticks_OSR_staging_10Hz_AmotechL5/11-185112.sbp` (and `.tag`) is the one I've been using

silverjam:
How hard would it be to use a unicode character and color it differently, something like https://www.fileformat.info/info/unicode/char/25cf/index.htm?  Layout looks a little odd on Windows:

Screenshot 2021-02-26 191718

notoriaga:
> How hard would it be to use a unicode character and color it differently, something like ●: https://www.fileformat.info/info/unicode/char/25cf/index.htm? Layout looks a little odd on Windows:
> 
> Screenshot 2021-02-26 191718

Hmm I tried that but couldn't find a way to change the text color afterwards. And the unicode chars with color don't seem to be supported

silverjam:
@notoriaga I got something working here: https://github.com/swift-nav/piksi_tools/pull/1227

silverjam:
> Hmm I tried that but couldn't find a way to change the text color afterwards. And the unicode chars with color don't seem to be supported

Hrmm, maybe I need to test on all platforms then, maybe it's something that works on Windows, but not Linux?

notoriaga:
> > Hmm I tried that but couldn't find a way to change the text color afterwards. And the unicode chars with color don't seem to be supported
> 
> Hrmm, maybe I need to test on all platforms then, maybe it's something that works on Windows, but not Linux?

What you have seems to work on linux too, I was trying to use extended unicode characters like - https://unicode-table.com/en/1F534/

silverjam (approved):

piksi_tools/console/fusion_engine_status.py
```diff
@@ -0,0 +1,153 @@
...
+    else:
+        return '* { color: grey; }'
+
+
+def check_flag(flag):
+    return lambda msg: (getattr(msg, flag) & 0b1111) == 0
+
+
+def stats_key(flag):
+    return f'stats_{flag}'
```
silverjam, line 78: We're using Python 3.5 for deploying currently 😬 -- so need to transform this to "old style" string interpolation

silverjam (approved):
lgtm, need to fix-up the f-strings then we're good

silverjam:
@denniszollo any feedback?

notoriaga:
CI seems to be running into some docker hub rate limit

silverjam:
@notoriaga still hitting some throttling but I re-kicked and it seems to have gotten past the docker pull step for now

denniszollo (approved):
Looks great!  I've done some limited testing and reviewed the logic and I think it behaves as expected and should help us understand that data is or is/not being used by fusion engine.

PR #1229, opened 2021-03-16
Fix INS Status printing as dec instead of hex

samvrlewis:
Looks like the INS status is currently printed in decimal instead of hex, which is a little confusing with the leading `0x` :)

Tested this by running the CI built console (`s3://swiftnav-artifacts/piksi_tools/v3.0.1.dev15+g45dda95/linux/swift_console_v3.0.1.dev15+g45dda95_linux.tar.gz`) and confirming the console showed the message in hex as expected.


denniszollo (approved):
Thanks for finding and fixing !

PR #1238, opened 2021-04-22
Add new settings for fusion engine for Piksi 2.5

samvrlewis:
Adds the antenna_offset_deviation and vehicle_frame_deviation settings
that are used in Piksi 2.5 to configure the fusion engine. Descriptions
and notes have been taken from the Starling user manual.

### DIFF
--- a/piksi_tools/console/settings.yaml
+++ b/piksi_tools/console/settings.yaml
@@ -2565,6 +2565,30 @@
   readonly: true
   Description: inertial navigation system build date
 
+- group: ins
+  name: antenna_offset_deviation
+  type: double
+  units: meters
+  default value: 0.05
+  readonly: False
+  Description: Standard deviation of antenna lever arm measurement.
+  Notes: |
+    Must be greater than 0.
+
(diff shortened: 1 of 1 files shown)


samvrlewis:
Let me know if there's anything else required here in the console.

Also worth noting that I've intentionally changed the default values for the deviations from the fusion engine (and so also from the documented starling manual defaults) as the default values gave pretty bad accuracy (it defaults to 1m for the antenna and 10 degrees for the vehicle frame, which seem like a lot to me). Maybe it makes more sense to be consistent with what starling/fusion-engine has though? I'm not sure how accurate these things are typically measured to.

denniszollo:
This looks good to me.  The default value here is just informational and should correspond to what happens if you press "reset to defaults", or the value that the memory defaults to where the setting was initialized via libsettings. I think it is fine to put something reasonable in there that differs from Starling defaults.

denniszollo (approved):
Approved as long as default value here corresponds to what we put in the sensor fusion pikelet for the default

PR #1239, opened 2021-04-23
Fix status bar flickering at 1hz update rate

samvrlewis:
Slightly increase the amount of time before data is timed out, from 1s
to 1.2s. This stops the console "flickering" back to showing "None" when
the output is set at 1Hz. This is assuming that 1Hz is the lowest output
rate possible.

Tested by running my Duro in time matched mode at `soln_freq=1` and `output_every_n_obs=1` and staring at the status bar for 5 minutes to confirm that there's no flickers of `Pos: None` or `Sats: 0`. Without this change, those would occasionally flicker to None/0.

Let me know if I misunderstood the intent of the last update time check though. As I understand it, it's to time out data if there hasn't been an update for a while.

### DIFF
--- a/piksi_tools/console/console.py
+++ b/piksi_tools/console/console.py
@@ -64,6 +64,11 @@ from piksi_tools.console.utils import (EMPTY_STR, pos_mode_dict,
 
 HEARTBEAT_CHECK_PERIOD_SECONDS = 1.2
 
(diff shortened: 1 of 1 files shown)


denniszollo (approved):
good change; I think more and more we will see 1 Hz operation.

PR #1242, opened 2021-05-05
Update default value for antenna deviation to string

denniszollo:
default values are a string in the schema


samvrlewis (approved):
My bad for getting this wrong, thanks for fixing!

jayvdb (approved):

PR #1258, opened 2021-09-10
Bumping PyInstaller and Qt package [DEVINFRA-548]

silverjam:
- Switch base Python to 3.6 for Linux and macOS so that we have access to more PyInstaller versions (and hopefully better support for later PySide2/PyQt5 versions)
- For macOS, switch to PySide2 5.15.2 and PyInstaller 4.2
  - PyQt 5.15.4 with PyInstaller 4.2 built with Xcode 8.3.3 crashes on start-up on macOS 10.14
  - PySide2 5.15.2 with PyInstall 4.2 (built with either Xcode 8.3.3 or 9.4) launches just fine, but the .dmg size increases to ~225mb from ~70mb
- Clean-up/remove ARM build stuff and clean-up some cruft from the setup.sh script
- Does not change Windows build, this stays at Python 3.5 and PyInstaller 3.6-- otherwise the build hangs and is eventually terminated by AppVeyor


switanis (commented):
I don't think I can provide any meaningful feedback here. This is not my domain.

john-michaelburke:
LGTM why the interest in more versions of pyinstaller? Just in general to keep it up to date? I guess py3.5 is EOL maybe that is why?

john-michaelburke (approved):

silverjam:
> 
> 
> LGTM why the interest in more versions of pyinstaller? Just in general to keep it up to date? I guess py3.5 is EOL maybe that is why?

Later versions of PyInstaller have better suport for PySide2, so in general if you want to use a recent version of this package you need to use a fairly recent version of PyInstaller.  Since Python 3.5 is EOL they dropped support for it, which prevents us from using an version of PyInstaller after they dropped support.

silverjam:
Tested new Linux build on Ubuntu 20.04 and Ubuntu 18.04.

PR #1265, opened 2021-10-11
workaround for macOS 11 white-on-white tabs [DEVINFRA-567]

silverjam:
Forces tab to be styled with macOS 11 blue accent color, which is what Qt defaults to on macOS 11.  Qt ignores changes in the accent color so this workaround doesn't attempt to match the accent color that's currently selected in the system preferences.

Appearance on macOS 11 (Big Sur)

Screen Shot 2021-10-12 at 9 49 56 PM

On macOS 10.15 (Catalina):

catalina_console

Linux and Windows look the same as before:

linux_console

Screenshot 2021-10-12 221455

### DIFF
--- a/piksi_tools/console/console.py
+++ b/piksi_tools/console/console.py
@@ -130,12 +130,19 @@ def get_args():
     return parser
 
 
-def is_macos11():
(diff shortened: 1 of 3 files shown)


john-michaelburke (approved):
Looks good to me! That windows GUI is going to look much better on the new console.

PR #1279, opened 2023-06-06
strip out GUI related code, optimize for factory use case

silverjam:
Jira: https://swift-nav.atlassian.net/browse/DEVINFRA-1101

As we try to modernize the `registar` and `piksi_provision_tools` code bases we're running into issues with aging Python dependencies.

This is maybe a somewhat controversial change since we're deleting all the GUI code.  However the rational is the following:
- The GUI code isn't maintained or used anymore, it's been replaced by Swift Toolbox: https://github.com/swift-nav/swift-toolbox
- The factory provisioning tools are the only remaining user (internally at least) and it does not need the GUI code
- Stripping away the GUI code makes it possible to run on more "exotic" hardware like Linux ARM which is showing up increasingly as a control machine platform for customers


denisng:
> makes it possible to run more "exotic" hardware like Linux ARM which is showing up increasingly as a control machine platform for customers

I'm also using RPis for Fac Infra where the computing power of a NUC is not required.

README.rst
```diff
@@ -7,67 +7,32 @@ Tools for the Piksi GNSS receiver
...
-  Latest console for `Piksi Multi <http://downloads.swiftnav.com/swift_console>`__
-
-  Older versions of console for use with `Piksi v2 <http://downloads.swiftnav.com/piksi_console>`__
-
-  For x86-64 Linux, tar.gz distributions are available for Swift Console `Piksi Multi <http://downloads.swiftnav.com/swift_console>`__
-
-To run the console GUI from the command line, install dependencies and run ``PYTHONPATH=. python piksi_tools/console/console.py``.
-
-For command line arguments, see `console.py <https://github.com/swift-nav/piksi_tools/blob/master/piksi_tools/console/console.py>`__
+  ``serial_link.py``, and ``settings.py`` - support Python 3.7 onward.
```
denisng, line 30: `libsettings` is also being recommended over `settings.py` in recent time

denisng (approved):
- [x] Builds wheels.
- [x] None of the things I use has been touched.

LGTM

README.rst
silverjam: Well, piksi_tools.settings uses libsettings, but we could probably move the settings module into libsettings and remove the one from here... the problem is the two libs are somewhat interwoven since libsettings would need to gain support to talk over serial/tcp (which, in example code, it circularly uses piksi_tools for...).

dgburr: swiftcar still uses `piksi_tools.settings`

achubbic-snav (approved):

switanis:
I think this is a good move (to remove obsolete GUI and make CLI working on ARM platforms).

switanis:
Recently Nexus Robotics were asking for:
> Basically we would like to be able to have a Swiftnav console and install piksi_tools via python on an ARM platform so we are able to edit the piksi settings via console or command line on the terminal.

switanis:
IMO no need for GUI. Just CLI.

switanis:
(of course customers are asking for everything)

silverjam:
The `piksi_tools` package works on ARM now:

image

silverjam:
> > > Recently Nexus Robotics were asking for:
> > > > Basically we would like to be able to have a Swiftnav console and install piksi_tools via python on an ARM platform so we are able to edit the piksi settings via console or command line on the terminal.
> > 
> > 
> > Any GUI on ARM is a pretty big lift, because we'll have to build PyQt6 ourselves since Qt does not provide a pre-built package for ARM: https://pypi.org/project/PySide6/6.5.1.1/#files
> > CLI tools are pretty easy.
> 
> https://packages.debian.org/bullseye/python3-pyside2.qtgui is available for the following architectures: amd64 arm64 armel armhf i386 mips64el mipsel ppc64el s390x

We're using PySide6, which doesn't look like it's available via Debian yet: https://packages.debian.org/search?keywords=python3-pyside.

PR #1281, opened 2023-06-08
Bump to python 3.9

achubbic-snav:
Effort to bring all piksi provisioning utilities up to python 3.9


pyproject.toml
```diff
@@ -13,7 +13,28 @@ packages = [
 ]
 
 [tool.poetry.dependencies]
-python = ">= 3.7"
+python = "^3.9"
+setuptools-scm = "3.3.3"
+numpy = "1.24.3"
+configparser = "^5.0.2"
+future = "0.17.1"
+futures = "3.1.1"
```
denisng, line 21: @silverjam, now that we have dropped support for Python 2, do we still need these backports?

Not critical to do in this change. I can open a ticket if we want to punt for now.

pyproject.toml
```diff
@@ -13,7 +13,28 @@ packages = [
...
+pep8 = "1.7.0"
+pyserial = "3.4"
+pylibftdi = "0.16.1.2"
+pyparsing = "2.2.0"
+pygments = "2.7.4"
+requests = "^2.20.0"
+six = "1.13.0"
```
denisng, line 28: @silverjam, ditto, dropping Python 2, do we still need this compatibility stuff

pyproject.toml
```diff
@@ -13,7 +13,28 @@ packages = [
 ]
 
 [tool.poetry.dependencies]
```
denisng, line 15: @silverjam, would having requirements defined here conflict with `setup.cfg`?

silverjam (approved):
This repo isn't using poetry for building it's wheel, sorry for the confusion around that.  We should just drop the `poetry` sections from the pyproject.toml.

silverjam (changes requested):
This repo isn't using poetry for building it's wheel, sorry for the confusion around that. We should just drop the poetry sections from the pyproject.toml.

silverjam:
@achubbic-snav please see https://github.com/swift-nav/piksi_tools/pull/1282

silverjam:
Close?

